PR #4, opened 2017-03-14
jmtaysom challenge 8 and 10

jmtaysom:


jmtaysom:
I realized that the name changes would remove the original files.

PR #6, opened 2017-03-24
challenge 11

atakume:
My first solution for challenges. 

Any suggestion will be welcome.

### DIFF
--- a/11/generators-template.py
+++ b/11/generators-template.py
@@ -14,22 +14,36 @@ $ for i in ../*/*py; do grep ^import $i|sed 's/import //g' ; done | sort | uniq 
    1 time
    1 datetime
 """
+import glob
+import re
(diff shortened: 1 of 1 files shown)


pybites:
Nice, thanks. For Counter you probably don't need to manually sort, because it has a most_common method. I will merge your solution on the community branch as 11/generators-atakume.py

pybites:
This PR thing is a bit of experiment, but it is here (note the branch in the URL), ok?
https://github.com/pybites/challenges/blob/community/11/generators-atakume.py

atakume:
Thank you for answer and putting my solution in repo. 
I sort manually to sort by both values and keys, I think most_common sorts by values only. 

Ok, I'll remember branch name for future requests.

PR #8, opened 2017-04-05
Challenge 13 - top directors 

atakume:
Suggestions welcome.

### DIFF
--- a/13/directors-atakume.py
+++ b/13/directors-atakume.py
@@ -1,0 +1,62 @@
+from collections import namedtuple
+from csv import DictReader
+from itertools import groupby
+
+MOVIE_DATA = 'movie_metadata.csv'
+NUM_TOP_DIRECTORS = 20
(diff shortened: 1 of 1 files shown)


pybites:
Cool, thanks, the merge worked. 

I need to check out groupby, seems adequate here. It seems the way to go as well if you go SQL/Pandas with this one. 

Maybe you could use enumerate for the print / counter bit?

atakume:
Thanks for comment and merge. 
I'll check out enumerate function and SQL package.

bbelderbos:
SQL I meant as a concept, not package, sorry. I saw somebody loading the data into sqlite and solving this with one SQL statement, very cool! 

I can't wait to redo this one in Pandas after studying it some more, I will let you know ...

atakume:
Oh this way, ok, thanks for explanation. 
With pandas it should be extra easy. Have fun with it.

PR #18, opened 2017-04-26
dante's Pcc16 upgraded

dseptem:
new PR with improvements to the code
thanks for the feedback!


dseptem:
added error 500 when quotes api fails 3 timtes

dseptem:
removed pywebview feature, while it was nice, it was unnecesary. everyone has a browser installed on their computer

pybites:
Perfect: shorter and cleaner code. I like the refresh button, I kept me engaged longer :)

Although for/while + else is frowned upon sometimes, it does seem to have its uses. I like how you expressed it here like abort if while completes (max tries)

Btw I will change 2 into 3, it seems to try only twice now (probably won't hit but just for correctness, as you said you wanted 3 times): 

def get_random_quote():
    tries = 0
    while tries < 3:  # need one more
        print('try {}'.format(tries+1))
        try:
            1/0
        except ZeroDivisionError:
            tries += 1
    else:
        print('Done')

get_random_quote()


Output:

pybites:
BTW for better UX you probably want to return an error to render_template instead of just the abort(500)

Now get_random_quote() returns 2 values, I would make that one. If success return a quote dict (keys: quote, author), if fail upon max tries return and empty dict (Pythonic 'falsy')

Then maybe do something like this?

    @app.route("/")
    def wisdom():
        quote = get_random_quote()  # easier if one ret value, because can check for False

        if not quote:
             error = 'Could not get quote, try again'
             return render_template('index.html', error=error), 500    # add conditional Jinja templ

        # here we are still good, follow happy path
        bio = get_author_bio(author)
        return render_template('index.html', quote=quote, bio=bio)   # smaller interface

PR #33, opened 2017-05-29
Pcc20 update

dseptem:
mostly refactors for better readability


dseptem:
Commited 0.5 version.
Tried hard to refactor for even more readability because the way I handled the actions didn't click with me until now.
I still think it needs improvement, though.

pybites:
cool, thanks

pybites:
Small edit I made on the community branch to be able to run the tests. Awesome btw you added tests. 


(venv) [bbelderb@macbook dseptem (community)]$ git diff
diff --git a/20/dseptem/README.rst b/20/dseptem/README.rst
index dec8e18..7947d35 100644
--- a/20/dseptem/README.rst
+++ b/20/dseptem/README.rst
@@ -41,4 +41,14 @@ It contains:

 * Two actors

-* One item
\ No newline at end of file
+* One item
+
+
+Note PyBites
+=========
+
+Requires Python 3.6 (uses syntax: {**self.actions, **self.moves} which did not work with < 3.6)
+
+To run tests:
+
+    $ python -m unittest tests.tests
diff --git a/20/dseptem/tests/tests.py b/20/dseptem/tests/tests.py
index 8e9b2da..bfe44dd 100644
--- a/20/dseptem/tests/tests.py
+++ b/20/dseptem/tests/tests.py
@@ -1,6 +1,8 @@
 import unittest
-from dseptem.rooms import boxes

+# https://stackoverflow.com/questions/1896918/running-unittest-with-typical-test-directory-structure
+import rooms
+from rooms import boxes

 class TestBoxes(unittest.TestCase):
     def setUp(self):

pybites:
Hey Dante, as promised ;)

Again great job here, I really like this as an example / use case for classes / OOP. 

Few things: 

- I like the clean (pep8) formatting and documentation. 

- object representation: __repr__ - this would be how to construct the object: 

    def __repr__(self):
        return self.door_description

I would do 

    def __repr__(self):
        return Room({!r}, {!r}, ...)'.format(self.arg1, self.arg2, ...)

(you could even replace Room by variable self.__class__.__name__)

example stdlib diff str vs repr:

>>> from datetime import datetime as dt
>>> now = dt.now()
>>> str(now)
'2017-06-04 21:19:15.652725'
>>> repr(now)
'datetime.datetime(2017, 6, 4, 21, 19, 15, 652725)'

- nice use of @property - I looked into them last week (PyBites article), really useful

- class Player(object), not needed in Python3, you can just do class Player: pass
(you use 3.6 as you have f-strings and this syntax: {**self.actions, **self.moves}, something I noted down in the readme)

- maybe you can breakdown this one in two properties? 
possible_actions_and_exits

because now you always throw away one variable: 
        actions, _ = self.location.possible_actions_and_exits
        _, exits = self.location.possible_actions_and_exits

why not: 
        actions = self.location.possible_actions
        exits = self.location.possible_exits

= cleaner. If I see multiple return values or "_and_" in variable names it is usually a hint to refactor

- you can shorten your list_to_dict:

    @staticmethod
    def list_to_dict(l, start):
        return {str(i): a for i, a in enumerate(l, start=start)}

>>> names = ['bob', 'julian', 'dante']
>>> list_to_dict(names, 0)
{'0': 'bob', '1': 'julian', '2': 'dante'}

>>> dict(enumerate(names, 0))
{0: 'bob', 1: 'julian', 2: 'dante'}

(= so short that you don't need a method)

- seems the 2 static methods (above list_to_dict and perform_action) are not called anywhere so you could take them out.

HTH
Bob

dseptem:
1) I confused __repr__ with __str__, I switched them to be proper.
2) Yeah, love that decorator.
3) I did it out of routine because I use Python 2.7 @ my job.
4) Are you looking at old commits? I need them together because of the
action numbering. Interacting and exiting are both two actions, but of a
different kind.
5) Good catch!
6) I do use perform_action: Line 89 of rooms.py

Thanks for the feedback! :)

PR #154, opened 2017-12-07
Solution C02 I fixed a bug that was mutating the draw list during validation.

yothenberg:
The _validation function accepts the draw list and then mutates it by removing letters. Subsequent uses of the draw list are then invalid. The unit tests didn't catch it. I fixed the bug by copying the list for local mutation and I added an assertion to check that draw is not mutated during validation.

Difficulty level (1-10): [2]
Estimated time spent (hours): [.25]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [yes, itertools.permutations]

### DIFF
--- a/02/game.py
+++ b/02/game.py
@@ -28,9 +28,10 @@ def input_word(draw):
 
 
 def _validation(word, draw):
+    unused_letters = list(draw) # copied because we are mutating the list
     for char in word.upper():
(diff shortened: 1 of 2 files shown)


bbelderbos:
Nice, thank you

pybites:
Good catch, thanks

PR #170, opened 2017-12-18
PCC45 TDD Based Fizz Buzz

mridubhatnagar:

**Difficulty level (1-10):** 
5
**Estimated time spent (hours):** 
2
**Completed (yes/no):** 
Yes
**I stretched my coding skills (if yes what did you learn?):** 

1) After a long break did unit testing.
2) Went through blogs/tutorials/videos for brushing up unit testing using Pytest.
3) Learned to use parametrize decorator in Pytest.
4) Good example, to learn when parametrized needs to be used and when it is not needed to
use it.
5) Learned that TDD one needs to write unit tests before the actual code. And not the other way round.
6) Iterated, re-iterated through the tests. Refactored the test code to this in the end.

**Other feedback (what can we improve?):** 

Use of Parametrize decorator is a very neat and clean way to write unit test cases. 
The scrabble game had many edge cases. Writing unit test cases for scrabble game 
challenge can be a good way to get a hang about how parametrize decorator needs to be 
used.

### DIFF
--- a/45/fizzy.py
+++ b/45/fizzy.py
@@ -1,6 +1,3 @@
-
-
-
 def fizzbuzz(i):
     """
     Returns numbers from 1 to 100.
--- a/45/fizzy.py
+++ b/45/fizzy.py
@@ -17,3 +17,4 @@ def fizzbuzz(i):
 
 for i in range(1,101):
     number = fizzbuzz(i)
+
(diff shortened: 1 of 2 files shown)


pybites:
Again nice log! Please also move the files to a username subdirectory as per challenge "Git setup" - https://codechalleng.es/challenges/45

PR #172, opened 2017-12-20
[WIP] PCC16 Query Your Favorite API

mridubhatnagar:
Tweaked around WordnikAPI to fetch one word of the day. The
random word generated is emailed every day to the recipient's 
mail address.

TODO:
Write Unit test cases

**Difficulty level (1-10):** 
7

**Estimated time spent (hours):** 
~10-11 ( to come up with a use case and solution)
Some more time to get a hang of cron job.

**Completed (yes/no):** 
Yes

**I stretched my coding skills (if yes what did you learn?)**: 

1) Coming up with a good use case to query the API.
2) Sending email notification using Python Script.
3) (Major Learning) Cron job. So, that it gives me one word daily in inbox.
4) Learned how to set tokens using environment variables instead of directly adding them in
script.
5) yay! I learned little bit about bash script, linux commands as well. :D 
6) SheBang
7) Utility of cron and anacron.

**Challenges**

1) Cron Job. 

As of now to implement this. It is assumed that system and wifi are up and running at the time when cron is scheduled. Not a good approach.

So, for systems that are not always up and running. Learnt that there is anacrontab in ubuntu. So, if my cron is scheduled at 4:00 am. But, in case at 4:00 my system was turned off. So, the time I switch on the system 10 mins after that I'll receive the email. 


16/build_vocabulary.py
```diff
@@ -0,0 +1,99 @@
...
+    msg = MIMEText(msg)
+    msg['Subject'] = 'Word Of The Day!'
+    msg['From'] = smtp_account
+    msg['To'] = mailto
+    smtp_server.send_message(msg)
+    smtp_server.quit()
+
+
+
+api_key = 'a2a73e7b926c924fad7001ca3111acd55af2ffabf50eb4ae5'
```
pybites: tuck that away in an env variable

16/build_vocabulary.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+
+    response = requests.get(url)
+    byte_response = response.content
+    unicode_response = byte_response.decode("utf-8")
+    parse_response(unicode_response)
+
+
+def parse_response(response):
+    """
```
pybites: nice docstrings

pybites:
nice work

16/build_vocabulary.py
mridubhatnagar: Thanks for the suggestion. Made changes. :)

pybites:
please move files in your username dir 16/username/files to avoid merge conflicts - https://codechalleng.es/challenges/16

pybites:
Nice documentation on your learning!
Please move your files in your username directory as per https://codechalleng.es/challenges/16

PR #189, opened 2017-12-29
PCC01 dylannakahodo

dylannakahodo:
Difficulty level (1-10): [1]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Relearned how to open a file and strip unnecessary characters.]

### DIFF
--- a/01/dylannakahodo/wordvalue.py
+++ b/01/dylannakahodo/wordvalue.py
@@ -9,13 +9,7 @@ def load_words():
 def calc_word_value(word):
     """Calculate the value of the word entered into function
     using imported constant mapping LETTER_SCORES"""
-    value = 0
-    for char in word:
-        try:
-            value += LETTER_SCORES[char.upper()]
-        except:
-            pass
-    return value
+    return sum(LETTER_SCORES.get(char.upper(), 0) for char in word)
 
 def max_word_value(word_list=load_words()):
     """Calculate the word with the max value, can receive a list


01/dylannakahodo/wordvalue.py
```diff
@@ -0,0 +1,33 @@
...
+
+def calc_word_value(word):
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    value = 0
+    for char in word:
+        try:
+            value += LETTER_SCORES[char.upper()]
+        except:
+            pass
```
pybites: "Explicit is better than implicit" - better to use a specific Exception on except (not except/pass), see https://pybit.es/py-mistakes.html

dylannakahodo: Thanks for reviewing! I realize this is a poor/lazy way of handling the key lookup in the dict after seeing other people's solutions.

pybites: right: use dict.get which defaults to None or a default value you give it. 

I (Bob) ended up with this one-liner for calc_word_value:     

return sum(LETTER_SCORES.get(char.upper(), 0) for char in word)

pybites: you can do this one here too now: https://codechalleng.es/bites/3/

dylannakahodo: Cool thanks! I'll check it out.

pybites:
cool

PR #197, opened 2018-01-07
Update tests to show correct count and update julian's twitter handle.

FabioRosado:
Hello, 

I've been trying for a while to figure out why we get 99 tweets when `NUM_TWEETS` is set to 100 or 100 when set to 101. I believe this might be an issue with tweepy, when we specify the `count` parameter to the `user_timeline()` it will stop one short (like when we use the `range()` function).

To counter that I've changed the tests to take `NUM_TWEETS` and reduce 1. Another way to do this would be update the solution version and add +1 to the count parameter, let me know if you'd rather have the solutions modified.

The code also fails when trying to get data from the handle `techmoneykids`, I made a guess and figured out that this was probably  Julian's old handle so I replaced it with the right one.

Please let me know what you think of this and if you need me to change something 👍

### DIFF
--- a/04/test_usertweets.py
+++ b/04/test_usertweets.py
@@ -10,7 +10,7 @@ HANDLE = 'pybites'
 MAX_ID = '819831370113351680'
 TWEETS = (
     """5 cool things you can do with itertools https://t.co/Nk4s3yL6zL #python""",
-    """How to create a nice-looking HTML page of your #Kindle book highlights (notes) https://t.co/HKFK7inhUa #python""",
+    """RT @guyfig: Just before 2016 is ending, here is a great answer from last week by Zuck.  @mkennedy @TalkPython #Python #devops https://t.co/…""",
 )
 USER = UserTweets(HANDLE, max_id=MAX_ID)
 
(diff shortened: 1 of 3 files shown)


bbelderbos:
Thanks Fábio, this must have been old work cause the tests call the API, my bad. This is a part you really would want to mock, see an example here: https://github.com/pybites/100DaysOfCode/blob/master/081/test_whotweeted.py - if you want to give it a try, otherwise I will do it at a later time. 

Cheers
Bob

Germione:
Hi Bob,
for what I see, the update has not been merge yet, so I have the same error as Fabio.
Have you planned to do this ? Do you need help ? Maybe you don't have time and i am gladly offer my help if you want :)

bbelderbos:
Hey @FabioRosado @Germione was gonna code this up but look at this nice PR: https://github.com/pybites/challenges/pull/221 - will verify and merge it now. Thanks for reminding me, these are busy months here.

pybites:
@Germione all good now, pull the community branch: your code passes with the new tests in place, nice.

@FabioRosado your code is not yet on the community branch, please PR the script (not the tests) and I will run the tests and merge it.

PR #198, opened 2018-01-08
PCC03 dinosaurabh

dinosaurabh:
Difficulty level (1-10): [5]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Learned about the feedparser module. Also learned about view objects and dict generators.]


pybites:
nice, loving feedparser, makes it so easy to parse rss feeds

PR #205, opened 2018-01-18
PCC02 deb17

deb17:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I have a different take on this problem. I did not use itertools.permutations at all. Instead I worked off the DICTIONARY.]
Other feedback (what can we improve?): [In the approach given in game-help.py, it may be worthwhile to add the permutations to a set instead of a list. As the draw can contain repeated letters, the permutations will have many duplicates. If a set is used, then the number of permutations to cross-check against the DICTIONARY will be considerably less. (Accordingly, the test case test_get_permutations_draw will need to be changed.)]

### DIFF
--- a/02/deb17/game.py
+++ b/02/deb17/game.py
@@ -1,0 +1,69 @@
+#!python3
+# Code Challenge 02 - Word Values Part II - a simple game
+# http://pybit.es/codechallenge02.html
+
+from data import DICTIONARY, LETTER_SCORES, POUCH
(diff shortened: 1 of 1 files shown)


pybites:
Interesting approach. Would need to run through it with the tests and time both solutions. Merging this to the community branch for now.

PR #207, opened 2018-01-23
PCC01 cactail

cactail:
Difficulty level (1-10): [2]
Estimated time spent (hours): [1.5]
Completed (yes/no): [yes]
stretched my coding skills (if yes what did you learn?): [Yes, I had to read the docs (again) about `key` parameter in max() built-in function]


pybites:
key is neat, finding people are familiar with it on sort(ed) but not so much on max/min

PR #234, opened 2018-03-24
PCC02 d3prof3t

d3prof3t:


pybites:
Nice!

pybites:
Saw you write about Python, cool. I have Generators open in a tab to check out later ...

d3prof3t:
Yea started it almost year back. Need to be regular on that. Let me know if you have any suggestions or something about them.

PR #241, opened 2018-04-05
PCC17 clamytoe

clamytoe:
Difficulty level (1-10): [5]
Estimated time spent (hours): [10]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Yes, this was when I was first learning click. I don't like the colors, so might take that out. Didn't test after modifying for this PR.]

### DIFF
--- a/17/clamytoe/main.py
+++ b/17/clamytoe/main.py
@@ -20,10 +20,14 @@ def main():
 
 
 @main.command()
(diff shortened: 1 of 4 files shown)


pybites:
I remember us (Bob) talking about this one a long time ago - nice work :)

click is an awesome module, I think I pip installed it so we can probably do a Bite on it as well

PR #251, opened 2018-05-02
PCC02 bmiraski

bmiraski:
Difficulty level (1-10): [4]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Plugged my way through itertools permutations, and raising my own errors that loop back. managed to do it without using the "help" file]


pybites:
itertools is useful, seems like a lot of good practice, thanks

PR #258, opened 2018-05-04
PCC05 bmiraski

bmiraski:
Difficulty level (1-10): [8]
Estimated time spent (hours): [12]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I got an answer. Ended up using Spacy on a rec from a friend. This gave answers that seemed to make more sense compared to what I was getting from Gensim. I was close with Gensim, and look at your solution to tweak what I had.  I just wasn't confident in the results I was getting back. Spacy probably would work even better with all the words (just stripped of some things like URLs), but this was a rough test (and not for PROD). Just glad this one is done. Need to keep reading on NLP and how it all works.]

### DIFF
--- a/05/bmiraski/.gitignore
+++ b/05/bmiraski/.gitignore
@@ -1,0 +1,3 @@
+data/*
+config.py
+similar_tweeters_gensim.py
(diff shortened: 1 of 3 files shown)


pybites:
nice work

PR #264, opened 2018-05-09
Pcc08

bmiraski:
Difficulty level (1-10): [5]
Estimated time spent (hours): [8]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Went for some more difficulty and built a full menu application that uses a sqlite3 database. This needs to be built with the initdb.py file before starting. I should have errors in place for all bad inputs by the users, and the output handles rooms that don't actually have inventory. Learned a bunch about sqlite3 (started some learning by doing a flask tutorial). I might go back and update for a Flask app at some point.]


pybites:
Nice! You can also try to use SqlAlchemy to abstract away the DB/SQL

PR #265, opened 2018-05-10
Hangman

bmiraski:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [A lot to think about with so many user inputs. You need to definitely keep in mind all the ways that someone could break things (or the initial setup could break things). Adjusted the user input string to allow for longer phrases, and eliminated all strings with numbers (sorry, 2001, and 12 Angry Men). Also decided to work in upper() instead of lower() to look nicer when printed on screen. ]

### DIFF
--- a/10/bmiraski/hangman.py
+++ b/10/bmiraski/hangman.py
@@ -1,0 +1,157 @@
+"""Play Hangman with movie titles."""
+
+from graphics import hang_graphics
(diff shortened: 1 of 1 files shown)


pybites:
Hangman is still one of our favorite game exercises to learn programming (and thinking about design)

PR #273, opened 2018-07-13
Put the template back

jallman112:
It seems an individual renamed and edited the template directly


bbelderbos:
Great catch, thanks!

PR #281, opened 2018-07-28
finished challenge 3

habereet:
Difficulty level (1-10): [4]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Learned about the difference between yield and return
Took me a while, in the -help version of the code to realize that calling dict(get_similarities() was causing a non-iterable type at one point. Maybe I should try, again, but with no-help?
Added the ability for the user to enter a similarity ratio to see different outputs at the end (no error checking)]


habereet:
Closing as I've updated the code to pass the unit tests.
Will open another pull-request.

Issue #220, opened 2018-02-11
Challange 04 test fails due to non-static data


alexjaw:
At this moment the required text for the last tweet appears instead already at item 97.
A solution is to mock the tweepy.API and inject static tweets. Will also dramatically decrease the test time (3s > 0.01s).

pybites:
I learned this skill > this challenge. I was just gonna give this a go (done it by now) but stumbled upon your PR (sorry, we have been busy), thanks a lot, will check verify and merge it now!

pybites:
Fixed recently

PR #292, opened 2018-08-28
PCC03

ansufei:
Difficulty level (1-10): 2
Estimated time spent (hours): 2
Completed (yes/no): yes
I stretched my coding skills (if yes what did you learn?): some new libraries/functions like SequenceMatcher. I'd be curious to know if there is a one-line solution for the get_similarities function.
Other feedback (what can we improve?): Did the rss file change or did I cheat? I had to modify the test file to complete the challenge (I had 102 distinct tags, the tag 'best practices' made the top ten, and there seemed to be 4 occurrences of the tag github (so I also modified the html file...)


bbelderbos:
Not sure, the tests use a static file: https://github.com/pybites/challenges/tree/master/03
Unlike PCC04 that used the live feed which we solved (https://github.com/pybites/challenges/issues/220)

pybites:
Would need to debug this with more time, feel free to discuss this on our #codechallenges channel on Slack (join via CodeChalleng.es)

pybites:
One-line solution for `get_similarities`? I would be amazed. That function needs some lines to do what it does, not sure how you can even shorten that without losing readability. Btw we found a nice performance improvement:

```
        # performance enhancements 1.992s -> 0.144s
        if pair[0][0] != pair[1][0]:
            continue
```

https://github.com/pybites/challenges/blob/solutions/03/tags.py

PR #295, opened 2018-08-31
Update generators.py

clamytoe:
What was I thinking before? Here's a refactored version.


pybites:
cool when you go back with more experience and refactor, no?

PR #298, opened 2018-09-01
PCC01 mh70cz

mh70cz:

Difficulty level (1-10): [2]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [90 % of time spent on git]

### DIFF
--- a/01/mh70cz/wordvalue.py
+++ b/01/mh70cz/wordvalue.py
@@ -1,0 +1,39 @@
+from data import DICTIONARY, LETTER_SCORES
+
+def load_words():
+    """Load dictionary into a list and return list"""
+    dl = []
+    with open(DICTIONARY) as f:
+        for line in f:
+            dl.append(line.strip())
+    return dl
+
+def calc_word_value(word):
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    score = 0
+    for l in word:
+        l = l.upper()
+        score += LETTER_SCORES.get(l, 0)
+    return score
+
+
+def max_word_value(words=None):
+    """Calculate the word with the max value, can receive a list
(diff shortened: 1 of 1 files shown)


01/mh70cz/wordvalue.py
```diff
@@ -0,0 +1,39 @@
+from data import DICTIONARY, LETTER_SCORES
+
+def load_words():
+    """Load dictionary into a list and return list"""
+    dl = []
```
pybites: You can also use `yield` to make this a generator, less code and performs faster

01/mh70cz/wordvalue.py
```diff
@@ -0,0 +1,39 @@
...
+    with open(DICTIONARY) as f:
+        for line in f:
+            dl.append(line.strip())
+    return dl
+
+def calc_word_value(word):
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    score = 0
+    for l in word:
```
pybites: `l`? I would use `word`

01/mh70cz/wordvalue.py
```diff
@@ -0,0 +1,39 @@
...
+        score += LETTER_SCORES.get(l, 0)
+    return score
+
+
+def max_word_value(words=None):
+    """Calculate the word with the max value, can receive a list
+    of words as arg, if none provided uses default DICTIONARY"""
+    max_val = 0
+    max_val_word = None
+    if words is None:
```
pybites: Nice use of `None` for argument checking. I also like this longer format solution as alternative to the shorter, but more cryptic `max` solution.

pybites:
Ping us here or on Slack what happened with `git`, I think we need to do a challenge git setup video soon ... Thanks

PR #302, opened 2018-09-04
PCC01 niilz_stroh

niilz:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [revisited my knowledge on reading in files, definitely understand a bit more about git now, learned that you can export/import static bindings (e.g. LETTER_SCORES), did not know you can also use the list comprehension-loop-style  on dictionaries (that was great to see). ]
Other feedback (what can we improve?): [I was not sure what to pass into the main function at the end. Don't know how unit testing works...yet. For now I wrote some print-statements ;)
This was fun! It was just my first challenge, but I really like that the tasks seem to be more complex than the little katas on codewars, without becoming to heavy on math. Which sometimes seem to be the only hurdle for me with some of them.]

### DIFF
--- a/01/niilz_stroh/wordvalue.py
+++ b/01/niilz_stroh/wordvalue.py
@@ -1,0 +1,28 @@
+from data import DICTIONARY, LETTER_SCORES
+
+def load_words():
+    """Load dictionary into a list and return list"""
+    with open(DICTIONARY, "r") as f:
+        words = f.read()
+        f.close()
+    return [word for word in words.split('\n')]
+
+def calc_word_value(word):
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    return sum([LETTER_SCORES[char] for char in word.upper() if word in load_words()])
+
+def max_word_value(wordList = load_words()):
+    """Calculate the word with the max value, can receive a list
+    of words as arg, if none provided uses default DICTIONARY"""
+    wordCounts = [(calc_word_value(word), word) for word in wordList]
+    return max(wordCounts)[1]
+
+if __name__ == "__main__":
+    print(load_words()[70:76])
+    #abashedly, abashedness, abashless, abashlessly, abashment, abasia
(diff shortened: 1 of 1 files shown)


pybites:
Great feedback, thanks!

01/niilz_stroh/wordvalue.py
```diff
@@ -0,0 +1,28 @@
+from data import DICTIONARY, LETTER_SCORES
+
+def load_words():
+    """Load dictionary into a list and return list"""
+    with open(DICTIONARY, "r") as f:
+        words = f.read()
+        f.close()
```
pybites: no need for `f.close()`, the with block (context manager) does that automatically

01/niilz_stroh/wordvalue.py
```diff
@@ -0,0 +1,28 @@
...
+    return [word for word in words.split('\n')]
+
+def calc_word_value(word):
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    return sum([LETTER_SCORES[char] for char in word.upper() if word in load_words()])
```
pybites: split over 2 lines (PEP E501)

01/niilz_stroh/wordvalue.py
```diff
@@ -0,0 +1,28 @@
...
+
+def max_word_value(wordList = load_words()):
+    """Calculate the word with the max value, can receive a list
+    of words as arg, if none provided uses default DICTIONARY"""
+    wordCounts = [(calc_word_value(word), word) for word in wordList]
+    return max(wordCounts)[1]
```
pybites: `max` takes an optional `key` argument so you can write: `max(words, key=calc_word_value)`, but the more verbose version might be more readable to somebody newer to the language. Just FYI

pybites:
Great work, keep calm and code in Python :)

PR #308, opened 2018-09-18
PCC01 rmackley

rmackley:
Difficulty level (1-10): [2]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I learned that it was possible to create a dictionary with a for loop and I learned how to use a unittest, sort of.]
Other feedback (what can we improve?): [The git instructions were a little unclear for me. I'm still not sure I did everything right. Edit on 9/17: the new instructions for setting up and using the git steps are much more clear. They definitely helped and the timing was great since I ended up having to do this twice.]


01/rmackley/wordvalue-template.py
```diff
@@ -0,0 +1,32 @@
...
+        for line in f:
+            dict_list.append(line.rstrip())
+    return (dict_list)
+
+def calc_word_value(test_str):
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    score = 0
+    for letter in test_str:
+        if letter not in [".",",","-","_",":",";"]:
```
pybites: if you want punctuation you can also use: `from string import punctuation`

01/rmackley/wordvalue.py
```diff
@@ -0,0 +1,32 @@
+from data import DICTIONARY, LETTER_SCORES
+
+def load_words(dict):
+    """Load dictionary into a list and return list"""
+    dict_list = []
```
pybites: I would call it `words`

01/rmackley/wordvalue.py
```diff
@@ -0,0 +1,32 @@
...
+        for line in f:
+            dict_list.append(line.rstrip())
+    return (dict_list)
+
+def calc_word_value(test_str):
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    score = 0
+    for letter in test_str:
+        if letter not in [".",",","-","_",":",";"]:
```
pybites: comment got deleted (?)
here you can use `punctuation` from the `string` module

pybites:
Nice work, glad the instructions made more sense after our adjustments yesterday!

PR #311, opened 2018-09-18
PCC51 - clamytoe (update)

clamytoe:
Cleaned up the code by running it through **black** and **isort**. Also used proper *SQL* syntax and refactored some of the code. The database is now recreated in seconds instead of over seven minutes!


bbelderbos:
resolved conflict by leaving the <<< PCC51 in, hope that works

PR #323, opened 2018-09-27
PCC01 vipinreyo

vipinreyo:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Helped to refresh the comprehensions.]
Other feedback (what can we improve?): [Nope, all good.]


01/vipinreyo/wordvalue.py
```diff
@@ -0,0 +1,30 @@
...
+def load_words():
+    """Load dictionary into a list and return list"""
+    with open(DICTIONARY, 'r') as df:
+        return df.read().split()
+
+
+def calc_word_value(word):
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    return sum([LETTER_SCORES[l.upper()] for l in word if l in string.ascii_letters])
```
pybites: I would break this up in 2 lines (pep8)

01/vipinreyo/wordvalue.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def max_word_value(words=None):
+    """Calculate the word with the max value, can receive a list
+    of words as arg, if none provided uses default DICTIONARY"""
+
+    if not words:
```
pybites: good, or `if words is None`

01/vipinreyo/wordvalue.py
```diff
@@ -0,0 +1,30 @@
...
+        words = load_words()
+
+    word_values = [(word, calc_word_value(word)) for word in words]
+    return sorted(word_values, key=lambda x: x[1], reverse=True)[0][0]
```
pybites: this seems more code than needed? what about using `max`?

pybites:
good work

01/vipinreyo/wordvalue.py
vipinreyo: Ok, got it. PEP 8 it is.

01/vipinreyo/wordvalue.py
vipinreyo: Yes, that's correct.

01/vipinreyo/wordvalue.py
vipinreyo: Yes, using max is pythonic.

PR #342, opened 2018-10-04
PCC53 vipinreyo

vipinreyo:
Difficulty level (1-10): [5]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Spotify Web apis are not so straight forward. It takes a bit of time to understand the type of Authorization approach to call the APIs.
I've started with Spotipy module and did a test run. Once all good, I've walked through the Spotipy code in github and coded my own Wrapper classes.]
Other feedback (what can we improve?): [One of the best challenges I've done as it matches with a real job requirement. Cheers.]

### DIFF
--- a/53/vipinreyo/spotify/Client.py
+++ b/53/vipinreyo/spotify/Client.py
@@ -1,0 +1,110 @@
+from Spotify import Spotify
+from Auth import Auth
+import os
+import sys
+
+"""Client script which calls the Spotify helper class methods to do the following.
+1. Query Spotify and grab a list of album names for any given artist. Bonus points if the script allows the user to
+    specify the artist!
+2. Create a playlist on your account if you don't already have one. Write a script that queries Spotify for the playlist
+    and returns all tracks/songs in the playlist.
+3. Write a script that obtains a list of the top tracks for an artist and see if any of those songs exist in your
+playlist.
+
+Pre-requisites:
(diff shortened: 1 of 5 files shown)


53/vipinreyo/spotify/Client.py
```diff
@@ -0,0 +1,110 @@
...
+            print('*' * FIFTY_TIMES)
+            print('My Play lists')
+            print('*' * FIFTY_TIMES)
+            for name in playlists:
+                print(name)
+            print()
+        else:
+            print("You don't have any public play lists.")
+
+    except Exception as ex:
```
pybites: I would make this try/except way smaller and use less indents (_extract to method_ refactoring type), remember zen: _Flat is better than nested._

53/vipinreyo/spotify/Spotify.py
```diff
@@ -0,0 +1,125 @@
...
+    def get_auth_headers(self):
+        """
+        Function to generate the header with the Authorization token
+        :return:
+        """
+        return {'Authorization': 'Bearer {0}'.format(self.auth_token)}
+
+    def _get(self, url, payload=None):
+        """
+        Helper method to call the Spotify APIs
```
pybites: nice docstrings

pybites:
great work here

PR #348, opened 2018-10-05
Added .mypy_cache/ to the list of ignored items to the .gitignore file

clamytoe:
Since I've been using mypy more frequently now, when adding files changed files, it keeps trying to add .mypy_cache/ as well. This will prevent it from happening to me in the future as well as to others.


bbelderbos:
great, thanks

PR #374, opened 2018-10-15
Pcc54

vipinreyo:
Difficulty level (1-10): [4]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Refreshed pypeclip and sqlite modules. 
PyQT5 documentation is evolving. Hence there are not much code available in the public domain to play around with, which is a constraint in designing GUIs for Python apps using QT.]
Other feedback (what can we improve?): [It would have better to give some references to Python GUI frameworks (which are being used and supported).]


pybites:
Nice to have a PyQT5 submission, great work

PR #396, opened 2018-10-27
Pcc01

therk:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Learned about rstrip and the use of dictionary get to set a default value.]
Other feedback (what can we improve?): [Best to reference data.py and dictionary.txt instead of copying them, so maybe put them into a module that can be referenced.]

### DIFF
--- a/01/therk/wordvalue.py
+++ b/01/therk/wordvalue.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python3
+
+from data import DICTIONARY, LETTER_SCORES
+
+
+def load_words():
+    """Load dictionary into a list and return list"""
+    return [line.rstrip('\n') for line in open(DICTIONARY)]
+
+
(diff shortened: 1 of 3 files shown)


pybites:
Yeah good idea, I have been soft linking them on the community branch but that is extra admin work, will look into this ...

01/therk/wordvalue.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
+
+from data import DICTIONARY, LETTER_SCORES
+
+
+def load_words():
+    """Load dictionary into a list and return list"""
+    with open(DICTIONARY, 'r') as file:
+        return [line.rstrip('\n') for line in file]
```
pybites: consider a generator here to reduce memory usage

PR #405, opened 2018-10-30
PCC42 danshorstein

danshorstein:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [a few extra re tricks]
Other feedback (what can we improve?): [the wording is a little confusing on the double word item, I assumed you were looking for just the single word that was duplicated (group(1)) based on the comment language in the function, but the test is looking for both words (group(0)). ]

### DIFF
--- a/42/danshorstein/regex.py
+++ b/42/danshorstein/regex.py
@@ -1,0 +1,46 @@
+import re
+
+
+def extract_course_times():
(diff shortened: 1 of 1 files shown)


danshorstein:
I’d change it from “double word” to “double words”

PR #410, opened 2018-10-31
PCC54 yashaslokesh

yashaslokesh:

Difficulty level (1-10): [4]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [Gained more experience representing datetime objects as strings in the format I want them, used error handling to make my code more concise, and found the useful string method splitlines() which gives a list of all strings separated by newlines. I can easily insert an element into the list and then reform the list into file contents. This is much easier than parsing through a string of all the file contents to find a certain location.]

### DIFF
--- a/54/yashaslokesh/copy_history.py
+++ b/54/yashaslokesh/copy_history.py
@@ -1,0 +1,60 @@
+from datetime import datetime
+from time import sleep
+
+import pyperclip
+
+''' pyperclip.copy() is queried every x seconds while script is running,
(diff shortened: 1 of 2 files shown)


pybites:
Great learning documentation, thanks. Nice error handling. I like `splitlines` but it is not lazy, is it? Only recently found out that _the file object is already a lazy generator of lines_ - see https://stackoverflow.com/a/519653

yashaslokesh:
Thanks for the tip, I'll remember to use it next time I'm reading in a file.

PR #429, opened 2018-11-20
Pcc42

rickhehe:
Difficulty level (1-10): [4]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Didn't know non-capture group can be used like that.]
Other feedback (what can we improve?): [add link to another awesome regex course for those who would like some further training on regex. https://youtu.be/VU60rEXaOXk]


pybites:
Thanks for the link, will add it to our to-watch list

pybites:
Yep non-capture groups are useful

PR #436, opened 2018-11-28
PCC13 - Highest Rated Movie Directors

zkan:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I learned about named tuple.]
Other feedback (what can we improve?): [I think the code `report = sorted(directors.items(), key=lambda x: float(x[0][1]), reverse=True)` in `test_directors.py` makes it difficult to guess what the output from `get_average_scores` would be. Therefore, it would be better if there is a comment explaining what the data structure of the output from `get_average_scores` is.]

### DIFF
--- a/13/zkan/directors.py
+++ b/13/zkan/directors.py
@@ -1,6 +1,7 @@
 import csv
 from collections import defaultdict, namedtuple
 
+
 MOVIE_DATA = 'movie_metadata.csv'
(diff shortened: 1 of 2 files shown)


pybites:
You want to make a PR updating the template code? Thanks

PR #445, opened 2018-12-04
'PCC02\ChronoHat'

ChronoHat:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Yes, I learned how to build a set using iterators]
Other feedback (what can we improve?): [It seemed like using a set as the output for permutations was the more efficient method. The unit test expects the more redundant list of all possible permutations, each of which gets checked against the dictionary]


pybites:
Interesting thanks, if you think the tests require a change, could you please open a PR against master?

PR #446, opened 2018-12-04
'PCC03_ChronoHat'

ChronoHat:
Difficulty level (1-10): [2]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Yes, I learned how to parse from xml documents. I also found a fun side effect of sets: as unordered collections, the order in which they return data is inconsistent and seems to be dependent on where that data is in memory. If it's important to display the information consistently or in the same order every time a program is run, either refrain from using a set or change it back to a list to sort it.]


pybites:
thanks for documenting your learning

PR #450, opened 2018-12-11
PCC11 - Generators for Fun and Profit

zkan:
Difficulty level (1-10): [4]
Completed (yes/no): [Yes]

### DIFF
--- a/11/zkan/generators.py
+++ b/11/zkan/generators.py
@@ -1,0 +1,50 @@
+"""
+Turn the following unix pipeline into Python code using generators
+
+$ for i in ../*/*py; do grep ^import $i|sed 's/import //g' ; done | sort | uniq -c | sort -nr
+   4 unittest
+   4 sys
+   3 re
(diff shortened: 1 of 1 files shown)


11/zkan/generators.py
```diff
@@ -0,0 +1,50 @@
...
+import re
+
+
+def gen_files(pat):
+    yield from glob.glob(pat)
+
+
+def gen_lines(files):
+    for each in files:
+        with open(each) as f:
```
pybites: have to double-check if we used it but starting (from memory) 3.3 you can use `yield from`

11/zkan/generators.py
```diff
@@ -0,0 +1,50 @@
...
+   1 time
+   1 datetime
+"""
+from collections import Counter
+import glob
+import re
+
+
+def gen_files(pat):
+    yield from glob.glob(pat)
```
pybites: nice

PR #465, opened 2019-01-04
Pcc01

leonardovega:
Difficulty level (1-10): [6]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I learned how to read a file and create a list from it. I practiced my recent git skills and was introduced to unit tests.]
Other feedback (what can we improve?): [I was a bit confused about whether or not to include the unit tests in my main file. Also I wasn't sure if we had to use the "unittest" package or if we could use regular "print()" commands.]

### DIFF
--- a/01/leonardovega/wordvalue.py
+++ b/01/leonardovega/wordvalue.py
@@ -1,0 +1,29 @@
+from data import DICTIONARY, LETTER_SCORES
+
+def load_words():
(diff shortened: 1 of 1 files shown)


pybites:
Sounds like great learning. Discussing the template files on Slack. `unittest` is builtin, do we need to add instructions how to invoke the tests?

PR #468, opened 2019-01-06
PCC11 bmcneill

bfmcneill:
Difficulty level (1-10): [5]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I learned about the power of generators when you need to pass lists, dicts, or sets around]

### DIFF
--- a/11/bfmcneill/generators-template.py
+++ b/11/bfmcneill/generators-template.py
@@ -1,0 +1,71 @@
+"""
+Turn the following unix pipeline into Python code using generators
+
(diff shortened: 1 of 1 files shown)


bfmcneill:
I could not figure out how to turn the get_result() function.  Any ideas?  It uses a default value for the generator.

pybites:
@bfmcneill you can check our solution here: https://github.com/pybites/challenges/blob/solutions/11/generators.py - we used `Counter`

PR #471, opened 2019-01-07
PCC58 bbelderbos

bbelderbos:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Nice challenge! I scraped talk python RSS feed (requests/ feedparser) getting the transcripts from GH, saved all this in a sqlite3 DB, also wrote a function to generate an HTML for consumption on my Kindle (after Save-to-PDF on Mac). Some details: `decode("utf-8")` to work around some transcripts being stored as `bytes` (required some debugging), `f'{episode_id:0>3}'` as alternative for `zfill`, cur.executemany is awesome as it works flawlessly with a generator of namedtuples, to get row dicts from sqlite use `conn.row_factory = sqlite3.Row`]


bbelderbos:
This was fun! 
screenshot 2019-01-07 at 15 17 35

PR #501, opened 2019-02-25
Pcc52

coderSoumya:

Difficulty level (1-10): []
Completed (yes/no): []


jcastle13:
The code runs fine but get the following error message.  I think the reason is because the 'say' command is supported in MAX OSX and not on Windows. 

`'say' is not recognized as an internal or external command`

PR #516, opened 2019-05-11
Pcc52

ppmadalin:

Difficulty level (1-10): [3]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [timedelta, datetime]
Other feedback (what can we improve?): [keep doing the good work]


jcastle13:
Hello @ppmadalin.  The code looks good.  The only suggestions I would make is to have a print statement at the top of the screen to let the user know that the code is a continuous timer.  You could set a print statement at the top of the while loop after line 13.  Also it would be nice to have an option, maybe a separate function, that you can ask the user for desired seconds and pass the argument in the countdown function.

PR #529, opened 2019-06-23
PCC04 jatinsinghal1921

jatinsinghal1921:
Difficulty level (1-10): [2]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I learned Tweepy API modules.]

### DIFF
--- a/04/jatinsinghal1921/usertweets.py
+++ b/04/jatinsinghal1921/usertweets.py
@@ -1,0 +1,85 @@
+from collections import namedtuple
+import csv
+import os
+
+import tweepy
+
+from config import CONSUMER_KEY, CONSUMER_SECRET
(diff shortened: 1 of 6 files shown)


04/jatinsinghal1921/usertweets.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+    def __len__(self):
+        """See http://pybit.es/python-data-model.html"""
+        return len(self._tweets)
+
+
+    def __getitem__(self, pos):
+        """See http://pybit.es/python-data-model.html"""
+        return (self._tweets)[pos]
```
feoh: Consider adding some type checks? As per https://docs.python.org/3/reference/datamodel.html#object.__getitem__ you can throw a TypeError here.

PR #531, opened 2019-06-25
PCC02 rguevara84

rguevara84:

Difficulty level (1-10): [2]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Learnt permutations and  validations]
Other feedback (what can we improve?): [good stuff]


pmayd:
It seems you have solved PCC01 and not 02, right? At least your folder says 01

pmayd:
Ah now I see you have files from challenge 01 and the `game.py` from challenge 02. So please separate your solutions for PCC 1 and PCC2. Make a MR for PCC 1 and commit only in folder 01 and for PCC 2 only in folder 02. When I understand it correctly, there is a version of `game-help.py` where you can add your functions directly from challenge 01

02/rguevara84/game.py
```diff
@@ -0,0 +1,102 @@
...
+    word = input().lower().strip()
+    if _validation(word, draw):
+        return word
+    else:
+        exit()
+    return word
+
+def _validation(word, draw):
+    """Validations: 1) only use letters of draw, 2) valid dictionary word"""
+    for letter in word:
```
pmayd: you can work with sets here with have nice methods to check intersection/overlap

02/rguevara84/game.py
```diff
@@ -0,0 +1,102 @@
...
+        if letter.upper() not in draw:
+            print("only use letters of draw")
+            raise(ValueError)
+        else:
+            state = True
+
+    if word not in DICTIONARY:
+        print ("Word not in dictionary")
```
pmayd: you don't need a `print` statement, each error class raised by `raise` can have a message that gets printed if the error is thrown: `raise ValueError("something went wrong...")`

02/rguevara84/game.py
```diff
@@ -0,0 +1,102 @@
...
+            pwords.append(word)
+    return pwords
+
+def _get_permutations_draw(draw):
+    """Helper for get_possible_dict_words to get all permutations of draw letters.
+    Hint: use itertools.permutations"""
+    # permutations = itertools.permutations(draw)
+    words = []
+    for x in range(1, len(draw)+1):
+        permutations = itertools.permutations(draw, x)
```
pmayd: why the extra for loop? you can use directly `for permutation in itertools.permutation(draw)` no need for the `x` here

pmayd: second, you dont get all permutations. You only draw the first len(draw)+1 permutations, but that are not all permutations. There are $n!$ permutations of a set of n different objects, otherwise n!/k! for each k of of repeated objects. For example "hello" does not have 5! = 120 permutations but 120 / 2 = 60, because the "l" letter comes twice.
However, 60 is not len("hello")+1 which would be 6. So you dont get all permutations

02/rguevara84/game.py
```diff
@@ -0,0 +1,102 @@
...
+# Below 2 functions pass through the same 'draw' argument (smell?).
+# Maybe you want to abstract this into a class?
+# get_possible_dict_words and _get_permutations_draw would be instance methods.
+# 'draw' would be set in the class constructor (__init__).
+def get_possible_dict_words(draw):
+    """Get all possible words from draw which are valid dictionary words.
+    Use the _get_permutations_draw helper and DICTIONARY constant"""
+    permutations = _get_permutations_draw(draw)
+    pwords = []
+    for word in permutations:
```
pmayd: use list comprehension `pwords = [word for word in permutations if word in DICTIONARY]`

PR #537, opened 2019-07-04
my solution

jbhonest:

Difficulty level (1-10): [5]
Estimated time spent (hours): [0.5]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [unittest]


01/wordvalue.py
```diff
@@ -0,0 +1,33 @@
+from data import DICTIONARY, LETTER_SCORES
+
+
+def load_words():
+    """Load dictionary into a list and return list"""
+    with open(DICTIONARY) as file:
+        words = file.read().strip().split('\n')
```
pmayd: do not use split with `\n` instead use `splitlines`. \n is not correct for all OS

01/wordvalue.py
```diff
@@ -0,0 +1,33 @@
...
+    return words
+
+
+def calc_word_value(word):
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    score = 0
+    for c in word:
+        if LETTER_SCORES.get(c.upper()):
```
pmayd: you dont need this check if you use the default return value 0 for get

01/wordvalue.py
```diff
@@ -0,0 +1,33 @@
...
+        of words as arg, if none provided uses default DICTIONARY"""
+    if words is None:
+        words = load_words()
+    max_score = 0
+    max_word = ''
+    for w in words:
+        t = calc_word_value(w)
+        if t > max_score:
+            max_score = t
+            max_word = w
```
pmayd: you can do it  this way but its more C++ style than Python. Better use a list comprehension and for each word create a tuple (value, word), and use the `max` function with the optional `key` argument

pmayd:
I cannot merge your PR as long as you do not comply with the user folder in the challenge folder. You have to create for each challenge a user folder where you put your files. Instead, you have put your files directly in 01 and not in 01/<user>

PR #542, opened 2019-07-14
Pcc30

feoh:
Difficulty level (1-10): [5]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [I refactored the logic in my code to be more testable by pulling it into methods, and then wrote unit tests, which found a bug in my logic that I then fixed, causing the tests to pass! Hear about this all the time but this is the first time it's happened to me! Super exciting :)]
Other feedback (what can we improve?): [I struggled a bit figuring out what to refactor because I had this idea in my head that only large programs with complex logic benefit from refactoring, but instead I chose the Pomodoro timer from a previous 100 days of code in Python lesson which only had one real piece of logic, but really surprised myself by getting a ton out of going through the exercise with that one small piece!]


30/feoh/pompom.py
```diff
@@ -0,0 +1,108 @@
...
+            completed_poms = completed_poms + 1
+            info(f"{completed_poms} pomodoros complete!")
+            print('\a')
+
+            break_time = decide_break_time(completed_poms, short_break_time, long_break_time)
+            info(f"Well done! Take a {break_time} minute break!")
+            time.sleep(break_time)
+            back_to_work()
+
+    except KeyboardInterrupt:
```
bbelderbos: there is quite some code under the exception, wonder if you would want to move that into a function

30/feoh/test_pompom.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+import pompom
+
+def test_decide_break_time():
```
bbelderbos: you could use `@pytest.mark.parametrize` here

30/feoh/pompom.py
```diff
@@ -0,0 +1,108 @@
...
+    pomodoro_seconds = 60 * pom_length
+    completed_poms = 0
+
+    try:
+        with open(poms_file,'r') as pfile:
+            Pomodoros = json.load(pfile)
+    except OSError:
+        info("No previously stored Pomodoros! Welcome!")
+
+    Pom={}
```
bbelderbos: flake8

30/feoh/pompom.py
```diff
@@ -0,0 +1,108 @@
...
+
+def back_to_work():
+    info("Back to work!")
+    print('\a')
+    print('\a')
+
+
+def decide_break_time(completed_poms, short_break_time, long_break_time):
+    """Determine whether or not the user has earned a break, and if so, how long?"""
+    break_reward_pom_count = 4
```
bbelderbos: move out into a constant at the top?

30/feoh/pompom.py
```diff
@@ -0,0 +1,108 @@
+import json
+import argparse
+import time
+import sys
+import readchar
+from logbook import RotatingFileHandler, info, notice, debug, StreamHandler
```
bbelderbos: pep8: order of imports

bbelderbos: cool, have not used `logbook` yet

30/feoh/pompom.py
bbelderbos: docstrings - nice

30/feoh/pompom.py
```diff
@@ -0,0 +1,108 @@
...
+
+def create_config_path():
+    config_path = Path.home() / ".config/pompom"
+    config_path.mkdir(parents=True, exist_ok=True)
+    return config_path
+
+
+def main(short_break_time, long_break_time):
+    config_path = create_config_path()
+    poms_file = config_path / "Pomodoro.json"
```
bbelderbos: pathlib, nice, I should use that more

bbelderbos:
Great job @feoh

30/feoh/pompom.py
feoh: It's really, *really* nice! I wish the documentation were better, but maybe I'll have the chance to contribute and improve that someday :)

For example, all you need is these two lines to set up an automatically rotated log file and simultaneously log to stdout:

```
    RotatingFileHandler("Pompom.log").push_application()
    StreamHandler(sys.stderr, level='INFO', bubble=True).push_application()
```

Pretty neat :)

PR #545, opened 2019-07-19
PCC62 clamytoe

clamytoe:

Difficulty level (1-10): [5]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [As always :)]
Other feedback (what can we improve?): [I found https://gender-api.com/ which would have given more accurate results, but they limit you to only 500 queries per month, which I would have blown through while testing. If I were to pursue this further I would use it on all of the names classified as unknown and use some sort of caching to look those up.]


pmayd:
As with @bbelderbos's PR, is a review needed here and if yes, whats our best-practice? As github is not supporting notebooks for review, I only see pulling the code and running the nb locally

bbelderbos:
Thanks @pmayd I could visualize the notebook in my browser now. Good stuff @clamytoe, going through some old PRs. Good trend for women speakers in Spain!

PR #546, opened 2019-07-19
Pcc62

serfer2:
Difficulty level (1-10): [5]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I did learn to use lxml library and building xpath for scraping]


62/serfer2/scraper.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        speakers = self.get_items_from_url()
+        for symbol in SPLIT_SYMBOLS:
+            speakers = self.split_names(speakers, symbol)
+        return self.filter_false_names(speakers)
+
+    def get_items_from_url(self) -> list:
+        """
+        Gets speaker's name list from a web page
+        Params:
```
pmayd: params are not up to date, method has no params beside self, which of course in a class is no real param

62/serfer2/scraper.py
```diff
@@ -0,0 +1,71 @@
...
+            url (str): webpage URL to crawl for.
+            xpath (str): XPATH to desired items to extract.
+        Returns:
+            list of str with speaker names or empty list if none
+        """
+        speakers = list()
+        response = requests.get(self._url)
+        if response.status_code == requests.codes.ok:
+            # import pdb; pdb.set_trace()
+            tree = html.fromstring(response.content)
```
pmayd: might wrap in a try block as fromstring can fail for ill-formed xml

62/serfer2/scraper.py
```diff
@@ -0,0 +1,71 @@
...
+        self._url = config['url']
+        self._xpath = config['xpath']
+
+    def get_names(self) -> list:
+        """
+        get all found names
+        """
+        speakers = self.get_items_from_url()
+        for symbol in SPLIT_SYMBOLS:
+            speakers = self.split_names(speakers, symbol)
```
pmayd: this seems not very Pythonic...by first reading over it I thought this can't be right as speakers is overwritten by split_names return value several times but because speakers is always the last returned value of this function this is correct in this context....but as I said it doesn't read nicely and is not obvious what is done here...so you pass the same input several times and split with different symbols. So if you need to split at multiple characters use `re.split` method which allows the definition of multiple characters with "or"

62/serfer2/scraper.py
```diff
@@ -0,0 +1,71 @@
...
+            speakers = tree.xpath(self._xpath)
+        return speakers
+
+    def split_names(self, names: list, symbol: str = ',') -> list:
+        """
+        Some times vairous speakers are published together
+        """
+        all_names = list()
+        for name in names:
+            all_names += [_n.strip() for _n in name.split(symbol)]
```
pmayd: no need for three lines use list comprehension

62/serfer2/scraper.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        A name is almost 2 words. A word must contain letters.
+        """
+        filtered_names = list()
+        for name in names:
+            are_words = True
+            words = name.split()
+            for word in words:
+                if word.strip().isdigit():
+                    are_words = False
```
pmayd: this is again not very Pythonic and reads much like JAVA or C. Instead of using a boolean flag, think about using the wonderful `all` and `any` methods in Python

62/serfer2/pycon_es.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+    names = dict()
+    with open(NAMES_CACHE_FNAME, 'r') as fi:
+        lines = fi.readlines()
+        for line in lines:
+            line = line.strip()
+            if not line:
+                continue
+            name, gender = line.split(';')
+            names[name] = gender
```
pmayd: can be combined in one list comprehension, even the assignment names = can be combined with list comprehension using `dict()` method with tuples. So something like `names = dict(((l.split(";")[0], l.split(";")[1]) for l in fi.readlines() if l))`

62/serfer2/pycon_es.py
```diff
@@ -0,0 +1,155 @@
...
+        for year, res in results.items():
+            female_pcnt = 0
+            male_qty = len(res.get('male', []))
+            female_qty = len(res.get('female', []))
+            qty = female_qty + male_qty
+            if qty > 0:
+                female_pcnt = float(female_qty * 100.0 / qty)
+                male_pcnt = float(male_qty * 100.0 / qty)
+            line = '%s,%d,%d,%.02f,%d,%.02f\n' %\
+                (year, qty, male_qty, male_pcnt, female_qty, female_pcnt)
```
pmayd: use Python's f-string notation to help the reader...

62/serfer2/pycon_es.py
```diff
@@ -0,0 +1,155 @@
...
+        for name, gender in names.items():
+            fo.write(f'{name};{gender}\n')
+
+
+def save_results(results):
+    """
+    saves resulting data to CSV
+    """
+    with open(RESULTS_FNAME, 'w') as fo:
+        fo.write('year,total,male,male %,female,female %\n')
```
pmayd: dont use 'male %' as a column name. It does work with pandas but only when passing the name as string. However, '%' is a very special char and should not be used in a column name that will be referenced often and maybe in different languages or packages. I suggest only use "_" as word delimiter and no special chars like "%". So much better would be your alternative 'male_pcnt' which by the way can be referenced in pandas with `df.male_pcnt'

62/serfer2/pycon_es.py
```diff
@@ -0,0 +1,155 @@
...
+    data_frame = pandas.read_csv(RESULTS_FNAME).sort_values('year')
+    data_frame.plot(
+        x='year',
+        y=['male %', 'female %'],
+        legend=True,
+        title='speaker\'s gender evolution (%)',
+        xticks=data_frame['year'].tolist(),
+        kind='bar'
+    )
+    plt.show()
```
pmayd: instead of calling `plt.show` you could consider returning the figure object which allows the caller to further change the graphical output

PR #551, opened 2019-07-27
Pcc52

abkosar:

Difficulty level (1-10): [5]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [Of course! This was Day 3 of 100 Days of Code in Python. I learned to use timedeltas.]

### DIFF
--- a/52/abkosar/timer.py
+++ b/52/abkosar/timer.py
@@ -1,0 +1,30 @@
+import argparse
+import time
+import os
+from datetime import datetime, timedelta
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Pomodoro Timer')
+    parser.add_argument('--duration',
+                        '-d',
+                        dest='duration',
+                        type=int,
+                        help='Duration of the timer in minutes',
+                        required=True)
+
+    args = parser.parse_args()
+    timer_duration = args.duration
+    timer = timedelta(minutes=timer_duration)
+    one_second = timedelta(seconds=1)
+
+    i = 0
+    print(timer.seconds)
+    while i < timer.seconds:
+        os.system("clear")
+        print(timer)
+        timer = timer - one_second
+        i += 1
+        time.sleep(1)
+
+    print("Your timer is completed!")


52/abkosar/timer.py
```diff
@@ -0,0 +1,30 @@
...
+    i = 0
+    print(timer.seconds)
+    while i < timer.seconds:
+        os.system("clear")
+        print(timer)
+        timer = timer - one_second
+        i += 1
+        time.sleep(1)
+
+    print("Your timer is completed!")
```
feoh: Super teeny nit - according to flake8/pep8 there should be a newline at end of file.

52/abkosar/timer.py
```diff
@@ -0,0 +1,30 @@
+import argparse
+import time
+import os
+from datetime import datetime, timedelta
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Pomodoro Timer')
```
feoh: You might consider refactoring your argument parsing into its own function.

52/abkosar/timer.py
```diff
@@ -0,0 +1,30 @@
...
+                        required=True)
+
+    args = parser.parse_args()
+    timer_duration = args.duration
+    timer = timedelta(minutes=timer_duration)
+    one_second = timedelta(seconds=1)
+
+    i = 0
+    print(timer.seconds)
+    while i < timer.seconds:
```
feoh: You might consider using something like a for loop and the range() built-in function here in order for your code to be more pythonic,

52/abkosar/timer.py
abkosar: Great advice, definitely will do that!

52/abkosar/timer.py
abkosar: True, will fix it.

52/abkosar/timer.py
abkosar: Should I fix these and do another PR?

PR #554, opened 2019-07-31
Pcc16

feoh:

Difficulty level (1-10): [3]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [This one got me comfortable using web APIS, and even more importantly taught me to validate my expectations before I implement. I came up with an awesome idea for a game based on the OMDB API that was not possible because though the API contains the required fields (ratings for the movies) there is no data in any of the entried!]
Other feedback (what can we improve?): [OMDB feels like a project on the wane to me. The data is extremely spotty and it seems like the poor developer / proprietor may be burned out or under water. Consider suggesting TMDB instead since it seems more full featured and better supported.]


16/feoh/randomtitlepicker.py
```diff
@@ -0,0 +1,41 @@
...
+# You'll need to create a file named config.py containing a line like: apikey="<your_api_key>"
+from config import apikey
+
+# Random OMDB title picker.
+
+
+def generate_random_imdb_id():
+    # This is the highest IMDB title id at time of writing. This stinks. Wish I had a better way to
+    # dynamically set this!
+    max_imdb_title_number = 9916754
```
bbelderbos: use a constant at the top of the module?

16/feoh/randomtitlepicker.py
```diff
@@ -0,0 +1,41 @@
+import requests
+import random
+# You'll need to create a file named config.py containing a line like: apikey="<your_api_key>"
+from config import apikey
```
bbelderbos: maybe commit a config-template.py file?

16/feoh/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pytest
+requests
```
bbelderbos: no omdb lib?

16/feoh/randomtitlepicker.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def get_random_movie() -> dict:
+    usable_movie = False
+    movie_json = {}
+
+    while True:
+        random_imdb_title_id = generate_random_imdb_id()
+        req = make_omdb_request(random_imdb_title_id)
+        movie_json: dict = req.json()
```
bbelderbos: so nice requests support the `json()` method no?

bbelderbos: huh what is that `: dict` syntax?

16/feoh/randomtitlepicker.py
```diff
@@ -0,0 +1,41 @@
...
+        if movie_json['Response'] == "True":
```
bbelderbos: what about checking `req.status_code == 200`?

bbelderbos: so if response != 200, can this cause an infinite loop? maybe have a max number of retries?

16/feoh/randomtitlepicker.py
```diff
@@ -0,0 +1,41 @@
...
+            break
+    return movie_json
+
+
+if __name__ == "__main__":
+    movie_json = get_random_movie()
+
+    for field in movie_json.keys():
+        if movie_json[field] != "N/A":
+            print(f"{field}: {movie_json[field]}")
```
bbelderbos: f-strings - noice

16/feoh/randomtitlepicker.py
bbelderbos: can you get this number from the API?

16/feoh/randomtitlepicker.py
```diff
@@ -0,0 +1,41 @@
...
+    # This is the highest IMDB title id at time of writing. This stinks. Wish I had a better way to
+    # dynamically set this!
+    max_imdb_title_number = 9916754
+    random.seed()
+    random_id = int(random.random() * max_imdb_title_number)
+    return f"tt{random_id}"
+
+
+def make_omdb_request(imdb_title_id):
+    movie_params = {'apikey': apikey, 'i': imdb_title_id}
```
bbelderbos: lowercase apikey it did not occur to me that it would come from config, use a constant here?

bbelderbos:
Great little project and learning, left some comments. HTH

16/feoh/randomtitlepicker.py
feoh: It's a type hint. Pycharm was trying to get me to use them. I've since decided I disagree with it :) IMO it makes the code less readable (to humans :) and doesn't gain you much.

16/feoh/randomtitlepicker.py
feoh: One would think so, right? As near as I can tell the answer is no, which is kind of underwhelmng :)

PR #578, opened 2019-09-25
Pcc13

bigarabuza:

Difficulty level (1-10): [8]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [I learned how to manipulate dictionaries, and named tuples. Also learned how to use __name__ = main(). I hadn't followed this convention before when writing scripts. It definitely helps adding structure to my programs and it's more modular with the use of functions.]
Other feedback (what can we improve?): [This was a good lesson and I really don't have much feedback on this one. Thanks!]


jcastle13:
Overall the code is great.  The only suggestion I would make is to look into making the program more robust with respect to the csv file.  Imagine if the csv file was in a different location (i.e. PWD) then the problem won't see that since it is looking one directory level above.  On possibility is to ask the user for the csv file or maybe place it in the same directory path as your code.  Other than that the code looks good.

bigarabuza:
That's a good point! I had the file pull in from the folder as that is how the directory was set up. I'll make a change to this soon. I definitely have a list of submissions/projects to refactor once I'm done w/ the 100 days of Code course.

PR #580, opened 2019-09-28
Pcc13

abidmunirmalik:
Difficulty level (1-10): [10]
Estimated time spent (hours): [20]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I really stretched myself to take care of this challenge, took me almost 2 days to finally resolved it.]
Other feedback (what can we improve?): [for a newbie like me, it make no sense to present the collections module's *Advanced* data structure(namedtuple, Counter, defaultdict)  without first introducing the *basic* data structure (list, tuple, dictionary). 
I can see the basic data structure followed by Advanced data structure in the course list, it should be other way around.]

### DIFF
--- a/13/abidmunirmalik/directors.py
+++ b/13/abidmunirmalik/directors.py
@@ -1,0 +1,46 @@
+import csv
+from collections import defaultdict, namedtuple
+
+MOVIE_DATA = 'movie_metadata.csv'
+NUM_TOP_DIRECTORS = 20
+MIN_MOVIES = 4
+MIN_YEAR = 1960
+
+Movie = namedtuple('Movie', 'title year score')
+
+
+def get_movies_by_director():
+    '''Extracts all movies from csv and stores them in a dictionary
+    where keys are directors, and values is a list of movies (named tuples)'''
+    pass
+
+
+def get_average_scores(directors):
+    '''Filter directors with < MIN_MOVIES and calculate averge score'''
+    pass
+
+
+def _calc_mean(movies):
+    '''Helper method to calculate mean of list of Movie namedtuples'''
+    pass
+
+
+def print_results(directors):
+    '''Print directors ordered by highest average rating. For each director
+    print his/her movies also ordered by highest rated movie.
+    See http://pybit.es/codechallenge13.html for example output'''
+    fmt_director_entry = '{counter}. {director:<52} {avg}'
+    fmt_movie_entry = '{year}] {title:<50} {score}'
+    sep_line = '-' * 60
+
+
+def main():
+    '''This is a template, feel free to structure your code differently.
+    We wrote some tests based on our solution: test_directors.py'''
+    directors = get_movies_by_director()
+    directors = get_average_scores(directors)
+    print_results(directors)
+
+
+if __name__ == '__main__':
+    main()
--- a/13/abidmunirmalik/directors.py
+++ b/13/abidmunirmalik/directors.py
@@ -1,45 +1,81 @@
 import csv
-from collections import defaultdict, namedtuple
+from collections import defaultdict, namedtuple, Counter
 
 MOVIE_DATA = 'movie_metadata.csv'
 NUM_TOP_DIRECTORS = 20
 MIN_MOVIES = 4
 MIN_YEAR = 1960
 
-Movie = namedtuple('Movie', 'title year score')
-
-
 def get_movies_by_director():
     '''Extracts all movies from csv and stores them in a dictionary
     where keys are directors, and values is a list of movies (named tuples)'''
-    pass
+    Movies  = namedtuple('Movies', ['title','year','score'])
+    movies_by_director = defaultdict(list)
+    with open(MOVIE_DATA) as csvfile:
+        reader = csv.DictReader(csvfile)
+        for rows in reader:
+            try:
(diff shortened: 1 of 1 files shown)


13/abidmunirmalik/directors.py
```diff
@@ -0,0 +1,82 @@
...
+
+MOVIE_DATA = 'movie_metadata.csv'
+NUM_TOP_DIRECTORS = 20
+MIN_MOVIES = 4
+MIN_YEAR = 1960
+
+def get_movies_by_director():
+    '''Extracts all movies from csv and stores them in a dictionary
+    where keys are directors, and values is a list of movies (named tuples)'''
+    Movies  = namedtuple('Movies', ['title','year','score'])
```
pmayd: namedtuple creates a class with the properties title year and score. So strictly spoken it is not "Movies" but "Movie", because each new Movie() is only one movie instance and not a collection of movies

13/abidmunirmalik/directors.py
```diff
@@ -0,0 +1,82 @@
...
+            movie = Movies(title=title, year=year, score=score)
+            movies_by_director[director].append(movie)     
+    return movies_by_director
+
+def filter_directors(directors):
+    directors_with_less_than_min_movies = []
+    for director,movies in directors.items():
+        d,m = director,len(movies)
+        if m < MIN_MOVIES:
+            directors_with_less_than_min_movies.append(d)
```
pmayd: this four lines can be simplified by using list comprehensions `directors_with_short_movies = [director for director, movies in directors.items() if len(movies) < MIN_MOVIES]`

13/abidmunirmalik/directors.py
```diff
@@ -0,0 +1,82 @@
...
+            except ValueError:
+                continue
+            movie = Movies(title=title, year=year, score=score)
+            movies_by_director[director].append(movie)     
+    return movies_by_director
+
+def filter_directors(directors):
+    directors_with_less_than_min_movies = []
+    for director,movies in directors.items():
+        d,m = director,len(movies)
```
pmayd: director is already available as variable, so "d" is not needed here and only confuses the reader :)

13/abidmunirmalik/directors.py
```diff
@@ -0,0 +1,82 @@
...
+        if m < MIN_MOVIES:
+            directors_with_less_than_min_movies.append(d)
+    for _ in directors_with_less_than_min_movies:
+        del directors[_]
```
pmayd: OK I understand now that you first build a list `directors_with_less_than_min_movies` to use it for deleting all entries in the original `directors` dictionary. That is a little more complicated than it has to be. You can use list comprehensions and create tuples, and the `dict` function will give you a dictionary. So 
`return dict([(director, movies) for director, movies in directors.items() if len(movies) < MIN_MOVIES])` is all you need in that function and gives you a dictionary with only directors that have less than MIN_MOVIES

13/abidmunirmalik/directors.py
```diff
@@ -0,0 +1,82 @@
...
+    return directors    
+
+    
+def get_average_scores(directors):
+    '''Filter directors with < MIN_MOVIES and calculate averge score'''
+    rating_ranks = defaultdict(list)
+    for director, movies in directors.items():
+        rating_ranks[director].append(_calc_mean(movies) )
```
pmayd: here you can again use list comprehensions to directly calculate the ratings

13/abidmunirmalik/directors.py
```diff
@@ -0,0 +1,82 @@
...
+    mylist = Counter(rating_ranks).most_common(20)
```
pmayd: you can do it your way or create an empty counter `c = Counter()` and use the update method. The second way avoids creating a dictionary `rating_ranks` because you directly update the counter

13/abidmunirmalik/directors.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def print_results(directors,final_list):
+    '''Print directors ordered by highest average rating. For each director
+    print his/her movies also ordered by highest rated movie.
+    See http://pybit.es/codechallenge13.html for example output'''
+    sep_line = '-' * 60
+    for i in range(len(final_list)):
+        counter = str((i+1)).zfill(2)+ '.'
+        d = final_list[i][0]
```
pmayd: this type of for loop is not common in Python, a more Pythonic way is to use `for d, m in final_list` or if you need the index for printing issues `for i, e in enumerate(final_list): d, av = e` because if e is an tuple with two elements you can use Python's magic unpacking and directly assign the two elements to two variables

13/abidmunirmalik/directors.py
```diff
@@ -0,0 +1,82 @@
...
+        av = final_list[i][1]
+        print("{} {} {}".format(counter,str(d).ljust(50),av))
+        print(sep_line)
+        print_Movie_list_by_director(directors,d)
+        print()
+
+def print_Movie_list_by_director(directors,director):
+    m = directors[director]
+    for i in range(len(m)):
```
pmayd: here again you can use `for i, m in enumerate(directors[director]):`

PR #582, opened 2019-09-30
Pcc52

BlaiseCosico:

Difficulty level (1-10): 3
Estimated time spent (hours): 2
Completed (yes/no): no
I stretched my coding skills (if yes what did you learn?): Datetime and timedelta

### DIFF
--- a/52/BlaiseCosico/pomodoro.py
+++ b/52/BlaiseCosico/pomodoro.py
@@ -1,0 +1,39 @@
+from datetime import datetime, timedelta
+import time as time
+
+def header():
+	print(f'\nWelcome to your very own Pomodoro Timer!\n\n'
+		f'--------------INSTRUCTIONS----------------\n'
+		f'Start the Pomodoro timer\n'
+		f'Work until the Pomodoro rings\n'
+		f'Take a short break (5 minutes)\n'
+		f'Keep on working, Pomodoro after Pomodoro, until the task at hand is finished.\n'
+		f'Every 3 Pomodoros take a longer break, (15–30 minutes).\n'
+		f'Enter how many sets do you want. (Each set has 3 Pomodoros)')
+    
+	set_loop(int(input()))
+
+def set_loop(ans):
+	while ans:
+		for i in range(3):
+			print('Time to work!')
+			start_timer(25)
+			print('Time to take a short break!')
+			start_timer(5)
+		print('Time to take a long break!')
+		start_timer(10)
+		ans-=1
+	print('Congratulations for finishing!')
+
+def start_timer(Minutes):
+	timer = timedelta(minutes=Minutes)
+	while timer:
+		print(f'Time remaining: {str(timer)}')
+		time.sleep(1)
+		timer -= timedelta(seconds=1)
+	print('Time is up!')
+
+	
+
+if __name__ == '__main__':
+	header()
--- a/52/BlaiseCosico/pomodoro.py
+++ b/52/BlaiseCosico/pomodoro.py
@@ -10,7 +10,6 @@ def header():
 		f'Keep on working, Pomodoro after Pomodoro, until the task at hand is finished.\n'
 		f'Every 3 Pomodoros take a longer break, (15–30 minutes).\n'
 		f'Enter how many sets do you want. (Each set has 3 Pomodoros)')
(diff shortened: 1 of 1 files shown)


52/BlaiseCosico/pomodoro.py
```diff
@@ -0,0 +1,38 @@
...
+
+def header():
+	print(f'\nWelcome to your very own Pomodoro Timer!\n\n'
+		f'--------------INSTRUCTIONS----------------\n'
+		f'Start the Pomodoro timer\n'
+		f'Work until the Pomodoro rings\n'
+		f'Take a short break (5 minutes)\n'
+		f'Keep on working, Pomodoro after Pomodoro, until the task at hand is finished.\n'
+		f'Every 3 Pomodoros take a longer break, (15–30 minutes).\n'
+		f'Enter how many sets do you want. (Each set has 3 Pomodoros)')
```
pmayd: No need for f-strings here, also it is not "wrong". Better suited is the trible string """ ... """ which also keeps the format. f-string is normally only used if you want to format or pass variables

52/BlaiseCosico/pomodoro.py
```diff
@@ -0,0 +1,38 @@
...
+			print('Time to work!')
+			start_timer(25)
+			print('Time to take a short break!')
+			start_timer(5)
+		print('Time to take a long break!')
+		start_timer(10)
+		ans-=1
+	print('Congratulations for finishing!')
+
+def start_timer(Minutes):
```
pmayd: your method not only starts a timer but also runs it and ends it ;) So name is misleading

52/BlaiseCosico/pomodoro.py
```diff
@@ -0,0 +1,38 @@
...
+		f'--------------INSTRUCTIONS----------------\n'
+		f'Start the Pomodoro timer\n'
+		f'Work until the Pomodoro rings\n'
+		f'Take a short break (5 minutes)\n'
+		f'Keep on working, Pomodoro after Pomodoro, until the task at hand is finished.\n'
+		f'Every 3 Pomodoros take a longer break, (15–30 minutes).\n'
+		f'Enter how many sets do you want. (Each set has 3 Pomodoros)')
+	set_loop(int(input()))
+
+def set_loop(ans):
```
pmayd: try to find better names for your parameters. It is not directly clear what `ans` means and which type it has.

52/BlaiseCosico/pomodoro.py
```diff
@@ -0,0 +1,38 @@
...
+	while ans:
+		for i in range(3):
```
pmayd: as you dont need the loop variable `i` you can replace `i` with `_` which is a placeholder in Python, so `for _ in range(3):` makes it clear to the reader that we don't need the variable

52/BlaiseCosico/pomodoro.py
```diff
@@ -0,0 +1,38 @@
...
+			print('Time to work!')
+			start_timer(25)
```
pmayd: why not let the user define the minutes?

52/BlaiseCosico/pomodoro.py
```diff
@@ -0,0 +1,38 @@
...
+		f'Start the Pomodoro timer\n'
+		f'Work until the Pomodoro rings\n'
+		f'Take a short break (5 minutes)\n'
+		f'Keep on working, Pomodoro after Pomodoro, until the task at hand is finished.\n'
+		f'Every 3 Pomodoros take a longer break, (15–30 minutes).\n'
+		f'Enter how many sets do you want. (Each set has 3 Pomodoros)')
+	set_loop(int(input()))
+
+def set_loop(ans):
+	while ans:
```
pmayd: `while` is technically wrong here because you know exactly how often to loop. As ans is an integer and gives the total amount of sets, you just would do a second loop, wouldn't you? No need for a while here and no need for `ans -= 1` in line 24

BlaiseCosico:
Wow thank you so much! I didn't realize someone would actually review my
code and give me pointers on how to improve it. And this was my first time
receiving a code review, so really, thank you! I'm sure to apply everything
that you said and work on being better in the future.

PR #585, opened 2019-10-01
PCC01

userinfamous:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Okay, starting from scratch took a bit of thinking but I have already completed word value before on one of the pybites so it wasn't that different.]
Other feedback (what can we improve?): [This is okay for an introduction.]


01/userinfamous/wordvalue.py
```diff
@@ -0,0 +1,33 @@
+from data import LETTER_SCORES, DICTIONARY
```
pmayd: you can also import from modules that are in the parent directory, so just use `from ..data import`. `.module` means Python is looking for module.py in the current directory and `..module` means Python is looking in the parent directory

01/userinfamous/wordvalue.py
```diff
@@ -0,0 +1,33 @@
+from data import LETTER_SCORES, DICTIONARY 
+import os
+
+#File path for Dictionary
+fname = os.path.join('../',DICTIONARY)
+
+#Wordlist from dictionary
+def load_words() -> list:
+    with open(fname,'r') as f:
+        wordlist = f.read().split()
```
pmayd: `read()` gives you the whole text in one string and `split()` depends on the default implementation, which is whitespace and line endings. But because the text is already with linebreaks, it is better to use `f.readlines` or to use the `splitlines()` method for strings

01/userinfamous/wordvalue.py
```diff
@@ -0,0 +1,33 @@
...
+        return wordlist
```
pmayd: return should be outside the with context. The lock is lifted, even for `return` inside a with block, but its good practice to not do this: https://stackoverflow.com/questions/9885217/in-python-if-i-return-inside-a-with-block-will-the-file-still-close

01/userinfamous/wordvalue.py
```diff
@@ -0,0 +1,33 @@
...
+
+def calc_word_value(word:str) -> int:
+    score = 0
+    for letter in word:
+        if letter.upper() in LETTER_SCORES:
+            score += LETTER_SCORES[letter.upper()]
```
pmayd: try to use list comprehensions: `score = sum(LETTER_SCORES.get(letter.upper(), 0) for letter in word)`. Normally we would have prooved `if letter.upper() in LETER_SCORES` but here we can use the convenient `get` method of dictionaries

01/userinfamous/wordvalue.py
```diff
@@ -0,0 +1,33 @@
...
+    return score
+
+def max_word_value(words=None):
+    wordvalue_pair = list()
+    #checking if given argument
+    if words == None:
+        wordlist = load_words()
```
pmayd: you can use `wordlist = words if words else load_words()`. This is one line in comparison to four and it uses the fact that Python evaluates None expressions to False

PR #592, opened 2019-10-02
Refactored PCC01

userinfamous:
+Refactoring code to be better, more readable and compact.
+Can't auto-merge, I want to get used to github better by merging.


01/userinfamous/wordvalue.py
```diff
@@ -1,33 +1,30 @@
...
 
 #File path for Dictionary
 fname = os.path.join('../',DICTIONARY)
 
 #Wordlist from dictionary
 def load_words() -> list:
     with open(fname,'r') as f:
-        wordlist = f.read().split()
-        return wordlist
+        wordlist = f.read().split('\n')
```
pmayd: do not use split but splitlines(), because `\n` is not always correct, windows uses `\r\n` for example

01/userinfamous/wordvalue.py
```diff
@@ -1,33 +1,30 @@
...
+    return wordlist[:-1]
```
pmayd: why the slice? Because the last element is empty? This is not obvious for a reder and seems strange

PR #594, opened 2019-10-02
PCC01 Anthlis

Anthlis:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I used the bite that I had previously completed for this challenge. This particular challenge was all about reviewing the GitHub PR workflow and command that I had some issues with before reaching this point. Hopefully, I have done it correctly. ]
Other feedback (what can we improve?): [Nothing. Good PR instructions, when followed properly! ]


01/Anthlis/wordvalue.py
```diff
@@ -0,0 +1,49 @@
...
+            word_list.append(line.strip())
+    return word_list
+
+
+def calc_word_value(word):
+    """given a word calculate its value using LETTER_SCORES"""
+    result = 0
+    # LETTER_SCORES is a dict so loop over and get the dict value each time +=
+    for letter in word:
+        result += LETTER_SCORES.get(letter.upper(), 0)
```
pmayd: use list comprehensions to further shorten your code. With list comprehensions you only need one line of code `result = [LETTER_SCORES.get(letter.upper(), 0) for letter in word]`

01/Anthlis/wordvalue.py
```diff
@@ -0,0 +1,49 @@
...
+def max_word_value(words=None):
+    """given a list of words return the word with the maximum word value"""
+    # use calc_word_value
+    max_value = 0
+    max_word = ""
+    for w in words:
+        word_value = calc_word_value(w)
+        if word_value > max_value:
+            max_value = word_value
+            max_word = w
```
pmayd: again you can use list comprehensions to transform each word in a tuple (word, value) or (value, word). Afterwrads, you can use the `sorted` function with the `key` argument to sort a list of tuples and max will then be the last entry. You can also use the `max` function that also support a key argument

Anthlis: Thanks for the feedback.

PR #597, opened 2019-10-02
Challenge Problem -1

Anuj-Devrani:


jcastle13:
Code looks good.  Would suggest for next PRs if you could fill out the description portion on how the code challenge was that would be great.  That helps for future challenges.  :)

PR #598, opened 2019-10-02
PCC03 GeoffRiley

GeoffRiley:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [New class, to me, `SequenceMatcher()`… some cunning functionality in there.]

### DIFF
--- a/03/GeoffRiley/tags.py
+++ b/03/GeoffRiley/tags.py
@@ -1,0 +1,41 @@
+import re
+from collections import Counter
+from difflib import SequenceMatcher
+from itertools import permutations
+
+TOP_NUMBER = 10
+RSS_FEED = 'rss.xml'
+SIMILAR = 0.87
+
+TAG_REGEX = re.compile(r'<category>([^<]+?)</category>')
+
+
+def get_tags() -> list:
+    """Find all tags in RSS_FEED.
+    Replace dash with whitespace."""
+    with open(RSS_FEED) as f:
+        return [t.replace('-', ' ').lower() for t in TAG_REGEX.findall(f.read())]
+
+
+def get_top_tags(tags):
+    """Get the TOP_NUMBER of most common tags"""
+    return list(Counter(tags).most_common(TOP_NUMBER))
+
+
+def get_similarities(tags):
+    """Find set of tags pairs with similarity ratio of > SIMILAR"""
+    return ((t, s) if t < s else (s, t) for t, s in permutations(set(tags), 2) if
+            t[0] == s[0] and SequenceMatcher(None, t, s).ratio() > SIMILAR)
+
+
+if __name__ == "__main__":
+    tags = get_tags()
+    top_tags = get_top_tags(tags)
+    print('* Top {} tags:'.format(TOP_NUMBER))
+    for tag, count in top_tags:
+        print('{:<20} {}'.format(tag, count))
+    similar_tags = dict(get_similarities(tags))
+    print()
+    print('* Similar tags:')
+    for singular, plural in similar_tags.items():
+        print('{:<20} {}'.format(singular, plural))


03/GeoffRiley/tags.py
```diff
@@ -0,0 +1,41 @@
+import re
+from collections import Counter
+from difflib import SequenceMatcher
+from itertools import permutations
+
+TOP_NUMBER = 10
+RSS_FEED = 'rss.xml'
+SIMILAR = 0.87
+
+TAG_REGEX = re.compile(r'<category>([^<]+?)</category>')
```
pmayd: if `[^<]+?` should match anything that is not a tag thats ok but if you use `(.*?)` this is more common as the ? means "not greedy version" and the closing </category> will mark the end. Without the ? you are right that the greedy version of the dot notation will match even the closing tag and beyond.

03/GeoffRiley/tags.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def get_tags() -> list:
+    """Find all tags in RSS_FEED.
+    Replace dash with whitespace."""
+    with open(RSS_FEED) as f:
+        return [t.replace('-', ' ').lower() for t in TAG_REGEX.findall(f.read())]
+
+
+def get_top_tags(tags):
```
pmayd: how many are the top tags? Maybe improve function name to show that the function returns a variable amount, use your constant as default: `def get_n_most_common_tags(tags, n=TOP_NUMBER):`

03/GeoffRiley/tags.py
```diff
@@ -0,0 +1,41 @@
...
+    """Get the TOP_NUMBER of most common tags"""
+    return list(Counter(tags).most_common(TOP_NUMBER))
+
+
+def get_similarities(tags):
```
pmayd: as you return a list of tag pairs, the function name is quite misleading in my opinion, because similarity sounds like a metric and I would expect to get the similarity metric of passed tags and not a list of tag pairs of two.

03/GeoffRiley/tags.py
```diff
@@ -0,0 +1,41 @@
+import re
+from collections import Counter
+from difflib import SequenceMatcher
+from itertools import permutations
+
+TOP_NUMBER = 10
+RSS_FEED = 'rss.xml'
+SIMILAR = 0.87
```
pmayd: again, not a good name. "SIMILAR" is a description and can be answered with yes or no and does not indicate a score or value. Better would be MIN_SIMILAR_VALUE or alike

03/GeoffRiley/tags.py
bbelderbos: I spent 30 min debugging a submission yesterday where there was just one month / int wrong due to a greedy/non-greedy problem :)

GeoffRiley: My reason for using `[^<]+?` rather than `.*?` is to emphasise that it is a second tag instance that is being avoided.  It also automatically excludes any empty tags.

03/GeoffRiley/tags.py
GeoffRiley: `get_top_tags(tags)` was provided as part of the boilerplate for the solution, as with the other routine names and constants.  Since I was working from the 'no-help' version of the template I assumed that these were to be taken as writ.

03/GeoffRiley/tags.py
GeoffRiley: This is another case of following the template :smile:

03/GeoffRiley/tags.py
GeoffRiley: Again, this is in the boilerplate, perhaps a note that these elements are just suggestions rather than directions?  It took me some time examining the tests provided to work out what `SIMILAR` was supposed to represent anyway.

03/GeoffRiley/tags.py
pmayd: Oh then I am very sorry and appologize for critizing it, or better for critizing you instead of the real culprit haha. That applies to all other comments as well. So there is not much then I can comment anyway ;)

GeoffRiley: No worries; I'm thick skinned enough to take much harsher peer review comments than this! :rofl:

PR #600, opened 2019-10-03
User reported issues with symlink to dictionary file in challenge 01 …

clamytoe:
…so copied file into here.


bbelderbos:
sounds good, thanks!

bbelderbos:
I tried to change base branch to master but that spoiled the diff, thinking about it you really only need this on community though no? all good then

bbelderbos:
awesome, so it might have been that way and it got deleted :) - thanks for fixing it!

PR #603, opened 2019-10-03
PCC02

userinfamous:
Difficulty level (1-10): [5]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [This one is a cheeky one... I learned a lot. Mostly my fault for getting stuck, see I forget to include re.IGNORECASE when using regex, see this is a very dumb mistake I made that cost me probably an hour.]
Other feedback (what can we improve?): [It's a good one, though the optional challenge seems like another tall tower to climb, why not split it into smaller bites? Because making a game putting it in a DB requires many many more hours.]

Sorry guys, this is my final pull request, I found out that the scramble took in more words than given this took me quite a bit to handle. But I handled it.


jcastle13:
Hello @userinfamous 

Was this PR for both PCC01 & PCC02?  If so, maybe creating two separate PR would be ideal.

userinfamous:
> Hello @userinfamous
> 
> Was this PR for both PCC01 & PCC02? If so, maybe creating two separate PR would be ideal.

I had  PCC01, but then for some reason, I changed some things, refactoring I guess but I forgot to switch branches so yeah.

jcastle13:
@userinfamous 

I removed PCC01 code.  I just noticed when I run the test_game.py that it doesn't pass due to the fact that some functions are missing inside of game.py.  Would it be possible to include the same functions in the test_game.py file?  This is good practice to have some form of unit testing since sometimes we can't always depend on manual testing.

userinfamous:
> @userinfamous
> 
> I removed PCC01 code. I just noticed when I run the test_game.py that it doesn't pass due to the fact that some functions are missing inside of game.py. Would it be possible to include the same functions in the test_game.py file? This is good practice to have some form of unit testing since sometimes we can't always depend on manual testing.

Yeah but like it's complicated when I try to make everything work, it affects all the other files and whatnot, overall it just makes a mess out of everything. For example, missing the dictionary file because for some reason the pointer doesn't work for me when importing functions from the parent.
So I just followed the format of everyone else after I manually run the tests.

02/userinfamous/game.py
```diff
@@ -0,0 +1,97 @@
...
+    """Calc the max value of a collection of words"""
+    return max(words, key=calc_word_value)
+
+# check if user input word can be made from the letters
+def input_check(word: str, pouch: list) -> bool:
+    matched = 0
+    for letter in word:
+        if letter.upper() in pouch:
+            matched += 1
+    return True if matched == len(word) else False
```
bbelderbos, line 29: `return matched == len(word)`

02/userinfamous/game.py
```diff
@@ -0,0 +1,97 @@
...
+    if re.search(word, ''.join(dictionary), re.IGNORECASE):
+        score = calc_word_value(word)
+        print('+ Word score is ', score)
+        return score
+    else:
+        print('- Word is not found')
+        return
+
+# function to reduce and find the best word
+def best_word(pouch: list, dictionary: set) -> int:
```
bbelderbos, line 43: there are parts that can be moved out of this function it seems to make it easier / more reusable

PR #608, opened 2019-10-05
Pcc45

Anthlis:
Difficulty level (1-10): [2]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I spent some time for this Code Challenge reviewing the #100daysofcode Pytest section and then set about writing my own fizzbuzz pytest code from scratch.
I also invoked pytest-cov from the CLI and experimented proving to myself that this only gives test coverage of the methods used and not the unique input parametrised tests from integer inputs 1 through to 100.
I use Mike Kennedy's design pattern of defining a main() under if __name__ == '__main__': and wondered if I should be testing this - how else can I get 100% test coverage, and is this important enough? (Feedback welcome!)]


bbelderbos:
@Anthlis funny to see this fizzbuzz exercise turned into a test Bite. Seems you need to delete the 31/ files for this one.

PR #613, opened 2019-10-06
PCC01 pwilson802

pwilson802:
Difficulty level (1-10): [3]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I was able to complete it fairly easily, however I'm not sure if its the most efficient solution.]

### DIFF
--- a/01/pwilson802/wordvalue.py
+++ b/01/pwilson802/wordvalue.py
@@ -17,7 +17,7 @@ def calc_word_value(word):
             score += LETTER_SCORES[letter.upper()]
     return score
 
-def max_word_value(words):
+def max_word_value(words = load_words()):
     """Calculate the word with the max value, can receive a list
     of words as arg, if none provided uses default DICTIONARY"""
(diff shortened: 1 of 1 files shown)


01/pwilson802/wordvalue.py
```diff
@@ -0,0 +1,30 @@
...
+def calc_word_value(word):
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    score = 0
+    for letter in word:
+        if letter.upper() in LETTER_SCORES:
+            score += LETTER_SCORES[letter.upper()]
+    return score
+
+def max_word_value(words):
```
schattencheg: You probably need to set default value for input parameter. 
Like this:
def max_word_value(words = load_words()):

pwilson802: Thanks for the feedback.  I made another commit with that change.

PR #614, opened 2019-10-06
PCC01 ronaldokun

ronaldokun:
Difficulty level (1-10): [4]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I haven't used yet the following method when reading a file:
`file.read().splitlines()`
But it worked perfectly for the following challenge, given it removed the newline character '\n' and returned a list]
Other feedback (what can we improve?): [I get this challenge is old but the instructions were very limited, I didn't know which files were supposed to be copied or not. I had to guess that.]

### DIFF
--- a/01/ronaldokun/wordvalue.py
+++ b/01/ronaldokun/wordvalue.py
@@ -1,0 +1,34 @@
+import re
+
+from data import DICTIONARY, LETTER_SCORES
+
+char = re.compile(r"[A-Z]{1}")  # I want to match individual UPPER CASE LETTERS
+
+
+def load_words() -> list:
+    """Load dictionary into a list and return list"""
+    with open(DICTIONARY, "r") as words:
+        return words.read().splitlines()
+
+
+def calc_word_value(word: str) -> int:
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    return sum(LETTER_SCORES[c] for c in char.findall(word.upper()))
+
+
+def max_word_value(list_of_words: list = None) -> str:
+    """Calculate the word with the max value, can receive a list
+    of words as arg, if none provided uses default DICTIONARY"""
+    if list_of_words is None:
+        list_of_words = load_words()
+    else:
+        list_of_words = list(list_of_words)
+
+    list_of_words.sort(key=calc_word_value, reverse=True)
+
+    return list_of_words[0]
+
+
+if __name__ == "__main__":
+    pass  # run unittests to validate


01/ronaldokun/wordvalue.py
```diff
@@ -0,0 +1,34 @@
...
+
+from data import DICTIONARY, LETTER_SCORES
+
+char = re.compile(r"[A-Z]{1}")  # I want to match individual UPPER CASE LETTERS
+
+
+def load_words() -> list:
+    """Load dictionary into a list and return list"""
+    with open(DICTIONARY, "r") as words:
+        return words.read().splitlines()
```
pmayd: do not use return within a context manager `with`. Athough the lock of the file is released in any case, you cannot easily catch exceptions

01/ronaldokun/wordvalue.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def calc_word_value(word: str) -> int:
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    return sum(LETTER_SCORES[c] for c in char.findall(word.upper()))
```
pmayd: always use `get` methods of dictionary if there is a chance that the passed key is not valid. So use `get(key, 0)` to present a valid default case that will not change the sum. Of course, here you use a regex to ensure that all found c are within the dictionary but you never know what happens and as you import DICTIONARY it might change some day

01/ronaldokun/wordvalue.py
```diff
@@ -0,0 +1,34 @@
+import re
+
+from data import DICTIONARY, LETTER_SCORES
+
+char = re.compile(r"[A-Z]{1}")  # I want to match individual UPPER CASE LETTERS
```
pmayd: Does this include for example german "ß"? :p Be careful with hardcoded allowed signed. It is much better to use the provided DICTIONARY with `get` method which will default to a default value for an invalid key. Dont try to predict all valid keys

01/ronaldokun/wordvalue.py
```diff
@@ -0,0 +1,34 @@
...
+    return sum(LETTER_SCORES[c] for c in char.findall(word.upper()))
+
+
+def max_word_value(list_of_words: list = None) -> str:
+    """Calculate the word with the max value, can receive a list
+    of words as arg, if none provided uses default DICTIONARY"""
+    if list_of_words is None:
+        list_of_words = load_words()
+    else:
+        list_of_words = list(list_of_words)
```
pmayd: use a one-liner `list_of_words = load_words if not list_of_words else list_of_words`

01/ronaldokun/wordvalue.py
```diff
@@ -0,0 +1,34 @@
...
+        return words.read().splitlines()
+
+
+def calc_word_value(word: str) -> int:
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    return sum(LETTER_SCORES[c] for c in char.findall(word.upper()))
+
+
+def max_word_value(list_of_words: list = None) -> str:
```
pmayd: you say that `list_of_words` is of type `list` but in line 26 you cast the variable to a list. So there is either a wrong type or an unnecessary list cast

01/ronaldokun/wordvalue.py
ronaldokun: I didn't know that. I've used this many times. I'll keep in mind from now on. Thank you!

01/ronaldokun/wordvalue.py
ronaldokun: At first I did that because in the test file was passed a tuple to test the function, so I put it this type cast to make it more general and cast iterables.

PR #616, opened 2019-10-07
PCC01 realnitinworks

realnitinworks:
Difficulty level (1-10): [1]
Completed (yes/no): [Yes]

### DIFF
--- a/01/realnitinworks/wordvalue.py
+++ b/01/realnitinworks/wordvalue.py
@@ -1,0 +1,39 @@
+from data import DICTIONARY, LETTER_SCORES
+
+
+def load_words():
+    """Load dictionary into a list and return list"""
+    with open(DICTIONARY) as f:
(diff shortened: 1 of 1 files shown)


01/realnitinworks/wordvalue.py
```diff
@@ -0,0 +1,39 @@
...
+    of words as arg, if none provided uses default DICTIONARY"""
+    if not words:
+        words = load_words()
+
+    word_values = {
+        word: calc_word_value(word)
+        for word in words
+    }
+
+    return max(word_values, key=word_values.get)
```
bbelderbos: could use `calc_word_value` as key here

realnitinworks: > could use `calc_word_value` as key here

Right. I missed that, Bob!

PR #622, opened 2019-10-08
PCC04 GeoffRiley

GeoffRiley:
Difficulty level (1-10): [4]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Uncovering the back office of Twitters API has been quite interesting.  The only thing that really complicated matters was the rigmarole of hoops that Twitter present in getting API keys.]
Other feedback (what can we improve?): [The API key application system is different… apps.twitter.com no longer exists and you have to register as a developer before you can go on to 'create' an app to get keys to use.  I put it down to being useful for my Open University Masters project on databases, but I'm not sure what a none education person would subscribe as.
Oh, also I had to increase the version of `tweepy` to 3.7.0 before it would run with Python 3.7.]

### DIFF
--- a/04/GeoffRiley/usertweets.py
+++ b/04/GeoffRiley/usertweets.py
@@ -1,0 +1,68 @@
+from datetime import datetime
+from collections import namedtuple
+import csv
+import os
+
+import tweepy
+
+from config import CONSUMER_KEY, CONSUMER_SECRET
+from config import ACCESS_TOKEN, ACCESS_SECRET
+
+DEST_DIR = 'data'
+EXT = 'csv'
+NUM_TWEETS = 100
+
+Tweet = namedtuple('Tweet', 'id_str created_at text')
+
+
+class UserTweets(object):
+    """TODOs:
+    - create a tweepy api interface
+    - get all tweets for passed in handle
+    - optionally get up until 'max_id' tweet id
+    - save tweets to csv file in data/ subdirectory
+    - implement len() an getitem() magic (dunder) methods"""
+
+    def __init__(self, handle, max_id=None):
+        self._handle = handle
+        self._save_file = f'{DEST_DIR}/{handle}.{EXT}'
+        self._max_id = max_id
+        self._auth = tweepy.OAuthHandler(CONSUMER_KEY, CONSUMER_SECRET)
+        self._auth.set_access_token(ACCESS_TOKEN, ACCESS_SECRET)
+        self._api = tweepy.API(self._auth)
+        self.read_tweets()
+        self.save()
+
+    def read_tweets(self):
+        twitterings = self._api.user_timeline(screen_name=self._handle, max_id=self._max_id, count=NUM_TWEETS)
+        self._twitterings = [Tweet(x.id_str, x.created_at, x.text) for x in twitterings]
+
+    def save(self):
+        if not os.path.exists(DEST_DIR):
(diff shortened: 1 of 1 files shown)


pybites:
Twitter API was easy at some time, shame they make it so difficult to get setup now :(

04/GeoffRiley/usertweets.py
```diff
@@ -0,0 +1,68 @@
...
+    - save tweets to csv file in data/ subdirectory
+    - implement len() an getitem() magic (dunder) methods"""
+
+    def __init__(self, handle, max_id=None):
+        self._handle = handle
+        self._save_file = f'{DEST_DIR}/{handle}.{EXT}'
+        self._max_id = max_id
+        self._auth = tweepy.OAuthHandler(CONSUMER_KEY, CONSUMER_SECRET)
+        self._auth.set_access_token(ACCESS_TOKEN, ACCESS_SECRET)
+        self._api = tweepy.API(self._auth)
```
pybites: could make a helper

04/GeoffRiley/usertweets.py
```diff
@@ -0,0 +1,68 @@
...
+        self.read_tweets()
+        self.save()
+
+    def read_tweets(self):
+        twitterings = self._api.user_timeline(screen_name=self._handle, max_id=self._max_id, count=NUM_TWEETS)
```
pybites: E501? not sure but I usually would use multiple lines

04/GeoffRiley/usertweets.py
```diff
@@ -0,0 +1,68 @@
...
+        self._twitterings = [Tweet(x.id_str, x.created_at, x.text) for x in twitterings]
+
+    def save(self):
+        if not os.path.exists(DEST_DIR):
+            os.makedirs(DEST_DIR)
+        if os.path.exists(self._save_file):
+            os.remove(self._save_file)
+        with open(self._save_file, 'w') as f:
```
pybites: abstract into helper?

04/GeoffRiley/usertweets.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @property
+    def output_file(self):
+        return self._save_file
+
+
+if __name__ == "__main__":
+
+    for handle in ('pybites', '_juliansequeira', 'bbelderbos'):
+        print('--- {} ---'.format(handle))
```
pybites: would be cool to make this a `__str__` dunder on the class I (Bob) think

pybites:
Very nice!

04/GeoffRiley/usertweets.py
GeoffRiley: Actually, I don't know why I let such a long line get through… I usually go through and deal with them before I push to git.  I use PyCharm so it commonly flags up if I've missed any too,.  Odd.

I totally agree with you though.

04/GeoffRiley/usertweets.py
GeoffRiley: Yes, I understand that concept. My reason for not doing so was that as it's a self contained object it wasn't going to be changing any of those elements after the initialisation.

However, one development I did consider was to make the authorisation and connection as a separate singleton class in order to isolate that part of the activity… but it seemed a little overkill for this example :)

PR #626, opened 2019-10-11
PCC04 pwilson802

pwilson802:
Difficulty level (1-10): [5]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [
- The tweepy module, I had never used this before. 
- How to return a generator from a function, seen it before but not used it.
- using Path from pathlib instead of the os module for creating a filename that can be used cross platform.
]


bbelderbos:
LGTM

PR #634, opened 2019-10-12
PCC01 rmachuca89

rmachuca89:

Difficulty level (1-10): [3]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [Configure unittest in VScode]
Other feedback (what can we improve?): 
Great practice projects and ways to keep the community engaged and learning
Only question is that I copied every file under my own folder. Not sure if only the main file had to be re-created (`wordvalue.py`)


jcastle13:
Hello @rmachuca89,

Yeah only the main file would need to be merged.  In this case the wordvalue.py.

PR #657, opened 2019-10-19
Pcc02

ronaldokun:
Difficulty level (1-10): [7]
Estimated time spent (hours): [8]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): Efficient iteration with the itertools and filter function.
Other feedback (what can we improve?): It's was not clear we must check for all permutations (NUM_LETTERS...1) to find the optimal word. I initially understood it was only NUM_LETTERS permutations, which is not the same.


ronaldokun:
>     1. You need to pull-request one challenge at a time. I.e. create 2 separated requests for 1st and 2nd challenges.
I know, but there were 2 minor commits in my repository from challenge 1 which entered the history 
>     2. Do not change additional files, like readme's and tests and do not commit these files.
I just removed the additional file in a new commit. I tried to revert the commit without success, just kept getting conflicts.

I don't know if this fixes it or which other additional action I have to take.

ronaldokun:
I undid the changes made in the previous commits with the recent one. What I need to do further?

PR #669, opened 2019-10-27
fixing tweepy exception invalid pagnination method

slipsnip:
Tweep.error('Invalid Pagnation Mode') caused inside TestUserTweets after patching only tweepy.API.user_timeline when you try to use the new Cursor method:

`cursor = tweepy.Cursor(api.user_timeline)`

Where api is the authorized api returned from `tweepy.API()`

### DIFF
--- a/04/test_usertweets.py
+++ b/04/test_usertweets.py
@@ -26,7 +26,7 @@ def read_csv(fname):
 class TestUserTweets(unittest.TestCase):
     def setUp(self):
         super().setUp()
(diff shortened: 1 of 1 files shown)


slipsnip:
Here is some supporting documentation of the problem https://yadi.sk/d/MAwRqryS0ocBpA
It is caused when you try to initialize a tweepy.Cursor object due to the unitest.mock.patch

bbelderbos:
Thanks @slipsnip for catching that!

PR #689, opened 2019-12-03
Pcc32

CodeitchNY:

Difficulty level (1-10): [5? relative and self-reported.]
Estimated time spent (hours): [6?]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [yes. surprised at some of the nuances of the site with selenium and working around them without reverting to time.sleep(). Might be a local issue, but had to be accommodated in the tests. Probably should have gone the class route, but what started simply got a little wet (IMHO, though part of it is the nature of the tests and selenium browser sessions).  A good chunk of time was spent before realizing that it worked headless MUCH better for the timings and focus and expected conditions (locally of course! ;) but it is better for it and made for a deeper review of pytest and selenium and not quick and dirty ]
Other feedback (what can we improve?): [I think you guys need to find a way to build this directly into your platform somehow. That would be neat. If feels "outside" ]


CodeitchNY:
whoa.. that must have been a brain fart. Closing my own pull request how the heck, 1300 files?

PR #697, opened 2020-01-15
PCC32 berubejd

berubejd:
Difficulty level (1-10): [6]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [This is the first time I've used Selenium]

### DIFF
--- a/32/berubejd/selenium_testing.py
+++ b/32/berubejd/selenium_testing.py
@@ -1,0 +1,234 @@
+#!/usr/bin/env python3.8
+
+from pathlib import Path
+from selenium import webdriver
+from selenium.webdriver.common.keys import Keys
+from time import sleep
+
+
+def setup(url=None):
+    if not url:
+        url = "http://pyplanet.herokuapp.com/"
+
+    print("\tStarting browser setup...")
+
+    chrome_driver = Path.cwd() / Path("chromedriver.exe")
+    driver = webdriver.Chrome(executable_path=chrome_driver)
+    driver.get(url)
+
+    return driver
+
+
+def test_homepage(driver):
+    """ Test homepage header and nav links """
+
+    print("\tPerforming anonymous user homepage testing...")
+
+    # Header
+    header = driver.find_element_by_xpath("/html/body/header/h1")
+    assert (
+        "PyBites 100 Days of Django" in header.text
+    ), "Incorrect header text or text missing"
+
+    # Login link
+    login_link = driver.find_element_by_xpath('//*[@id="login"]/a[1]')
+    assert "Login" in login_link.text, "Incorrect login link or link missing"
+
+    # Home link
+    home_link = driver.find_element_by_xpath('//*[@id="login"]/a[2]')
+    assert "Home" in home_link.text, "Incorrect home link or link missing"
+
+
+def test_sharer_app(driver):
+    """ Test sharer app link and generated table """
+
+    print("\tPerforming anonymous user App testing...")
+
+    # Navigate to the Sharer App
+    driver.find_element_by_xpath("/html/body/main/ul/li/a").click()
+
+    # Collect table header and links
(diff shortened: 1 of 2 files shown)


32/berubejd/selenium_testing.py
```diff
@@ -0,0 +1,234 @@
...
+from time import sleep
+
+
+def setup(url=None):
+    if not url:
+        url = "http://pyplanet.herokuapp.com/"
+
+    print("\tStarting browser setup...")
+
+    chrome_driver = Path.cwd() / Path("chromedriver.exe")
```
feoh: You might choose to make this cross platform (Path already handles the directory part!) by doing a check for Windows and tacking the .exe on the end if you're on that platform.

I feel like there should be something to do this in Pathlib already but I couldn't find it.

32/berubejd/selenium_testing.py
```diff
@@ -0,0 +1,234 @@
...
+
+    # Navigate to the Sharer App
+    driver.find_element_by_xpath("/html/body/main/ul/li/a").click()
+
+    # Collect table header and links
+    table_header = driver.find_element_by_xpath("/html/body/main/table/thead")
+    assert "Title" in table_header.text, "Incorrect table header"
+
+    table_links = driver.find_elements_by_xpath("/html/body/main/table/tbody/tr/td/a")
+    assert len(table_links) == 100, "Incorrect number of table rows"
```
feoh: You might wish to make 1000 into a CONSTANT at the top so it's easily updated should this value need to change.

32/berubejd/selenium_testing.py
```diff
@@ -0,0 +1,234 @@
...
+    # Find and save the first article in the table
+    table_link = driver.find_element_by_xpath("/html/body/main/table/tbody/tr[1]/td/a")
+    link_title = table_link.text
+
+    # Navigate to that article
+    sleep(2)
+
+    table_link.click()
+
+    # Check that this is the correct article by comparing titles
```
feoh: Might this be unnecessary if you used one of the Selenium Python driver's Wait functions?

Docs at:
https://selenium-python.readthedocs.io/waits.html

32/berubejd/selenium_testing.py
berubejd: This is a good point.  It is complicated slightly because I am running a Windows executable in Ubuntu under WSL.  Let me see if I can work something out.

32/berubejd/selenium_testing.py
berubejd: Agreed.  This is a better pattern.

32/berubejd/selenium_testing.py
berubejd: I tried both implicit and explicit waits but explicit waits probably deserve some additional research into the best way to make this work rather than hardcoded sleep statements.

32/berubejd/selenium_testing.py
berubejd: Applied a quick change that does a file search and, utilizing Path, should work on both Linux and Windows, including my WSL setup.

32/berubejd/selenium_testing.py
berubejd: Moved to a constant.

32/berubejd/selenium_testing.py
berubejd: I retested a vast number of wait related changes.  Strangely, in every case, the page was entirely rendered and all elements were being properly extracted from the DOM.  Input continued to fail without the sleep statements, however.

I implemented javascript input calls that worked without issue.  This may or may not be related to a bug somewhere in Selenium or the Chrome webdriver but, although there are a number of complaints, neither project is looking to address is.  The change I implemented is the only one recommended by the community that didn't require while loops and sleep statements.

I learned a lot more about Selenium waits, however. :)

PR #726, opened 2020-02-26
Pcc11

Zaubeerer:

Difficulty level (1-10): [7]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [generators, grep, yield from, islice]
Other feedback (what can we improve?): [maybe a short introduction to grep]

### DIFF
--- a/11/Zaubeerer/generators.py
+++ b/11/Zaubeerer/generators.py
@@ -31,13 +31,13 @@ def gen_lines(files):
     
     for file in files:
         with open(file) as f:
-            yield f.readline()
+            yield f.readlines()
 
 def gen_grep(lines, pattern):
     """Analyze whether lines match a certain pattern"""
     for line in lines:
         if line.startswith(pattern):
-            yield line[len(pattern):]
+            yield line[len(pattern):-1]
 
 def gen_count(lines):
     """Count number of occurrences of specific patterns"""
--- a/11/Zaubeerer/generators.py
+++ b/11/Zaubeerer/generators.py
@@ -51,7 +51,15 @@ if __name__ == "__main__":
     line_pattern = "import "
  
     files = gen_files(file_pattern)
+    print(files)
     lines = gen_lines(files)
+    print(lines)
     lines_of_interest = gen_grep(lines, line_pattern)
+    print(lines_of_interest)
     counted_lines = gen_count(lines_of_interest)
 
+    print(counted_lines)
+
+    print(*counted_lines)
+
+    [print(key, value) for key, value in counted_lines]
--- a/11/Zaubeerer/generators.py
+++ b/11/Zaubeerer/generators.py
@@ -20,35 +20,39 @@ import re
 
 import collections
 
+
 def gen_files(pat):
     """Finds all files"""
 
     yield from glob.glob(pat)
 
+
 def gen_lines(files):
     """Finds all lines"""
-    
+
     for file in files:
         with open(file) as f:
             yield from f.readlines()
+
 
 def gen_grep(lines, pattern):
     """Analyze whether lines match a certain pattern"""
     for line in lines:
         if line.startswith(pattern):
-            yield line[len(pattern):-1]
+            yield line[len(pattern) : -1]
+
 
 def gen_count(lines):
     """Count number of occurrences of specific patterns"""
 
     yield collections.Counter(line for line in lines)
-        
+
 
 if __name__ == "__main__":
     # call the generators, passing one to the other
-    file_pattern = '../*/*.py'
+    file_pattern = "../*/*.py"
     line_pattern = "import "
- 
+
     files = gen_files(file_pattern)
     lines = gen_lines(files)
     lines_of_interest = gen_grep(lines, line_pattern)
--- a/11/Zaubeerer/generators.py
+++ b/11/Zaubeerer/generators.py
@@ -15,10 +15,10 @@ $ for i in ../*/*py; do grep ^import $i|sed 's/import //g' ; done | sort | uniq 
    1 datetime
 """
 
-import glob, os
-import re
-
 import collections
+import glob
+import os
+import re
 
 
 def gen_files(pat):
--- a/11/Zaubeerer/generators.py
+++ b/11/Zaubeerer/generators.py
@@ -51,15 +51,6 @@ if __name__ == "__main__":
     line_pattern = "import "
  
     files = gen_files(file_pattern)
-    print(files)
     lines = gen_lines(files)
-    print(lines)
     lines_of_interest = gen_grep(lines, line_pattern)
-    print(lines_of_interest)
     counted_lines = gen_count(lines_of_interest)
-
-    print(counted_lines)
-
-    print(*counted_lines)
-
-    [print(key, value) for key, value in counted_lines]
--- a/11/Zaubeerer/generators.py
+++ b/11/Zaubeerer/generators.py
@@ -23,13 +23,11 @@ import re
 
 def gen_files(pat):
     """Finds all files"""
-
     yield from glob.glob(pat)
 
 
 def gen_lines(files):
     """Finds all lines"""
-
     for file in files:
         with open(file) as f:
             yield from f.readlines()
@@ -44,7 +42,6 @@ def gen_grep(lines, pattern):
 
 def gen_count(lines):
     """Count number of occurrences of specific patterns"""
-
     yield collections.Counter(line for line in lines)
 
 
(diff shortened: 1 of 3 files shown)


Zaubeerer:
@bbelderbos, the tests are now passing. Does it make sense to merge both `generators.py` and `test_generators.py`?

11/Zaubeerer/test_generators.py
```diff
@@ -0,0 +1,57 @@
...
+
+    expected_counts = [Counter({'generators': 2, 'numpy': 1})]
+
+    assert list(generators.gen_count(lines)) == expected_counts
+
+def test_integrated():
+
+    expected_counter = Counter({'re': 8, 'glob': 6, 'collections': 2, 'sys': 1, 'glob, os': 1, 'generators': 1, 'ast': 1})
+
+    files = generators.gen_files("../*/*.py")
```
bbelderbos: this did not work on my system, can you make the path a constant / load from env?

11/Zaubeerer/test_generators.py
```diff
@@ -0,0 +1,57 @@
...
+
+    lines = generators.gen_lines(files)
+    lines_of_interest = generators.gen_grep(lines, LINE_PATTERN)
+    counted_lines = generators.gen_count(lines_of_interest)
```
bbelderbos: did you get a chance to use `islice` or are you happy consuming the whole generator in this test?

11/Zaubeerer/test_generators.py
```diff
@@ -0,0 +1,57 @@
+import generators
+from collections import Counter
+from itertools import islice
+
+FILE_PATTERN = '*.py'
+LINE_PATTERN = "import "
+
+def test_gen_files():
+    expected_files = ["test_generators.py", "generators.py"]
+
```
bbelderbos: I think you can get rid of the empty lines as these are short functions

11/Zaubeerer/generators.py
```diff
@@ -0,0 +1,55 @@
...
+"""
+
+import glob, os
+import re
+
+import collections
+
+def gen_files(pat):
+    """Finds all files"""
+
```
bbelderbos: remove blank lines inside efunctions

11/Zaubeerer/generators.py
```diff
@@ -0,0 +1,55 @@
...
+    yield from glob.glob(pat)
+
```
bbelderbos: did you run pep8? should be 2 newlines between functions

11/Zaubeerer/generators.py
```diff
@@ -0,0 +1,55 @@
...
+def gen_grep(lines, pattern):
+    """Analyze whether lines match a certain pattern"""
+    for line in lines:
+        if line.startswith(pattern):
+            yield line[len(pattern):-1]
+
+def gen_count(lines):
+    """Count number of occurrences of specific patterns"""
+
+    yield collections.Counter(line for line in lines)
```
bbelderbos: yield from?

11/Zaubeerer/generators.py
```diff
@@ -0,0 +1,55 @@
...
+        
+
+if __name__ == "__main__":
+    # call the generators, passing one to the other
+    file_pattern = '../*/*.py'
```
bbelderbos: got me files of my homedir, please make configurable

11/Zaubeerer/test_generators.py
```diff
@@ -0,0 +1,57 @@
+import generators
+from collections import Counter
+from itertools import islice
+
+FILE_PATTERN = '*.py'
+LINE_PATTERN = "import "
```
bbelderbos: nice

11/Zaubeerer/test_generators.py
```diff
@@ -0,0 +1,57 @@
...
+
+    expected_counter = Counter({'re': 8, 'glob': 6, 'collections': 2, 'sys': 1, 'glob, os': 1, 'generators': 1, 'ast': 1})
+
+    files = generators.gen_files("../*/*.py")
+
+    lines = generators.gen_lines(files)
+    lines_of_interest = generators.gen_grep(lines, LINE_PATTERN)
+    counted_lines = generators.gen_count(lines_of_interest)
+
+    assert list(counted_lines)[0] == expected_counter
```
bbelderbos: actual == expected, nice order

11/Zaubeerer/test_generators.py
```diff
@@ -0,0 +1,57 @@
+import generators
```
bbelderbos: pep8 import order

11/Zaubeerer/generators.py
Zaubeerer: Just added black formatting. Sorry for sending unformatted code 🙈

11/Zaubeerer/test_generators.py
Zaubeerer: I just added 'sorting on save'

11/Zaubeerer/test_generators.py
Zaubeerer: Thanks!

11/Zaubeerer/test_generators.py
Zaubeerer: What do you mean?

I used `../*/*py` as described in the challenge:

`$ for i in ../*/*py; do grep ^import $i|sed 's/import //g' ; done | sort | uniq -c | sort -nr`

11/Zaubeerer/generators.py
Zaubeerer: If I use `yield from` I get only the keys of the `Counter`.

I wrote the test function such, that `gen_count()` is expected to return a `Counter`.

Is that wrong?

11/Zaubeerer/test_generators.py
Zaubeerer: I have not yet completely understood, when one should use yield.

I thought it's for very big generators to avoid unnecessarily big tests?

How would you use it here?

11/Zaubeerer/test_generators.py
Zaubeerer: I have removed some empty lines, does it now suit your expectations? :)

11/Zaubeerer/generators.py
Zaubeerer: See above, could you please clarify?

bbelderbos: ok yeah that is on us then, we should have better defined.

I mean all good when working in current directory, but not very portable so ideally want to configure the path here. All good for this exercise though.

11/Zaubeerer/generators.py
bbelderbos: This yields the whole thing at once. If that passes tests then that's on us (I would need to check), but if the collection is super large I think you want to yield them one by one, e.g.

```
def gen_count(lines):
    return (f'{cnt[1]:>7} {cnt[0]}' for cnt in Counter(lines).most_common())
```

or

```
    cnt = Counter(lines)  # Counter works on iterables and is not memory heavy
    yield from cnt.most_common()  # most_common sorts them descending
```

(examples from community branch of this repo!)

11/Zaubeerer/generators.py
bbelderbos: pep does not complain, I would only use them in functions that are > 5 lines, not needed here

11/Zaubeerer/test_generators.py
bbelderbos:
```
def gen():
    i = 0
    while True:
        i += 1
        yield i

assert list(gen()) == [1,2] # oops, hangs!
```

So let's only check 10 items:

```
from itertools import islice

g = gen()
assert list(islice(g, 2, 5)) == [3, 4, 5]
assert list(islice(g, 0, 3)) == [6, 7, 8]
```

works!

11/Zaubeerer/test_generators.py
bbelderbos: yeah ok on us, ok here, in real life put it in config

Zaubeerer: Alright. :)

config then refers to an external config file or a variable / constant at the file head?

11/Zaubeerer/test_generators.py
Zaubeerer: OK, got it! 

Thanks!

11/Zaubeerer/generators.py
Zaubeerer: Yeah, I also just checked it and found this appealing. 

There are also one-line solutions using regexs, neat!

PR #768, opened 2020-09-19
Solution to Challenge 01 by Ritam Dey

RitamDey:
Signed-off-by: Ritam Dey <deyritam27031999@gmail.com>

Difficulty level (1-10): [1]
Estimated time spent (hours): [.5]
Completed (yes/no): [yes]


01/RitamDey/wordvalue.py
```diff
@@ -0,0 +1,33 @@
+from data import DICTIONARY, LETTER_SCORES
+
+def load_words():
+    """Load dictionary into a list and return list"""
+    words = []
```
pmayd, line 5: no need to initialize first as long as you dont have any conditional statements or a try except block, here words gets overwritten or open() throws an error, so never will be empty (besides the document is empty but then the list comprehension will be empty)

01/RitamDey/wordvalue.py
```diff
@@ -0,0 +1,33 @@
...
+        words = [word.strip("\n") for word in dictionary]
+    return words
+
+def calc_word_value(word: str):
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    total_sum = 0
+    for char in word:
+        if char.upper() in LETTER_SCORES:
+            total_sum += LETTER_SCORES[char.upper()]
```
pmayd, line 16: you can use a list comprehension like in line 7 to merge these four lines into one with something like `sum(LETTER_SCORES.get(char.upper(), 0) for char in word)`. Notice my use of `get` which is another very very Pythonic solution for not writing `if key in dict do...else` each time. It provides a default value if the key is not found.

01/RitamDey/wordvalue.py
```diff
@@ -0,0 +1,33 @@
...
+def max_word_value(words=load_words()):
+    """Calculate the word with the max value, can receive a list
+    of words as arg, if none provided uses default DICTIONARY"""
+    word_score = -1
+    word_max = ""
+    for word in words:
+        score = calc_word_value(word)
+        if score > word_score:
+            word_score = score
+            word_max = word
```
pmayd, line 29: this is a ok solution but more in a world of C++ or Java, not in Python. We use functional programming and list comprehensions as much as possible. So think about what you want to achieve here: return the word with the max score. So this is the same as sorting all words by its score and returning the first or last entry, depending on whether you sort ascending or descending. So how can you sort in Python? Easy, use `sorted(list)`. But how can we tell the function to sort not by the words itself but their score? Easy, use the key param `sorted(words, key=calc_word_value)`. This will sort all words by their word value. All that is left to do is to get the word with max value, so take the last entry:
`sorted(words, key=calc_word_value)[-1]` and you have your solution in 1 intsead of 7 lines.

pmayd (approved):
A ok solution, can be improved in more Pythonic ways, but you are on the right track!

PR #769, opened 2020-09-20
Initial commit PyBites Challenge#52 PomodoroTimer Simple version

reachsrisris:

Difficulty level (1-10): []
Completed (yes/no): []

### DIFF
--- a/52/srisris/pomodoro/timer.py
+++ b/52/srisris/pomodoro/timer.py
@@ -1,0 +1,16 @@
+""" PCC52 - Create your own Pomodoro Timer 
+    Instructions + submit work: https://codechalleng.es/challenges/52"""
+
+from datetime import datetime, timedelta
+
+now = datetime.now()
+interval = timedelta(minutes=1)
+end_time = now + interval
+print(now)
+print(end_time)
+
+stopFlag = False
+while stopFlag == False:  
+    if datetime.now() >= end_time:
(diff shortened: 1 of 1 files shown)


reachsrisris:
Initial commit PyBites Challenge#52 PomodoroTimer Simple version

52/srisris/pomodoro/timer.py
```diff
@@ -0,0 +1,16 @@
...
+
+from datetime import datetime, timedelta
+
+now = datetime.now()
+interval = timedelta(minutes=1)
+end_time = now + interval
+print(now)
+print(end_time)
+
+stopFlag = False
```
pmayd, line 12: Here you need a boolean var to track the status. This is fine but a more Pythonic way would be to use `while True:` and just break out or the loop as soon as the end condition has arrived, namely `if datetime.now() >= end_time: break`

52/srisris/pomodoro/timer.py
```diff
@@ -0,0 +1,16 @@
...
+while stopFlag == False:
```
pmayd, line 13: while will tests much much faster then what you need for a timer, so it would be a good idea to use some `time.sleep` or equal methods to delay the code execution up to the next reasonable interval, like 1 second.

pmayd (approved):
Good first version, left some comments to improve

PR #773, opened 2020-10-03
Pcc13

souhailshiv:

Difficulty level (1-10): []
Completed (yes/no): []


pmayd (changes requested):
Please one PR per challenge, you have also commited files for challenge 52

souhailshiv:
> Please one PR per challenge, you have also commited files for challenge 52

what's do you mean I'm sorry I'm still a beginner in this website

pmayd:
No problem!  We expect you to open a pull request for each challenge you have solved. However when you look at the files tab you have pushed two files, one in folder 13 and one in 52. So you are handing in two different challenges. One pull request should only contain files in a single folder

13/souhailshiv/Highest_Rated_Movie_Directors.py
```diff
@@ -0,0 +1,83 @@
...
+        cnt[director] += len(movies)
+        c = dict(cnt)
+    for k, v in c.items():
+        if v >= MIN_MOVIES:
+            newDict[k] = v
+    avrege = dict()
+
+    for k in newDict.keys():
+        moviescores = list()
+        for item in directors[str(k)]:
```
pmayd, line 50: Item is not a good name, try to be more specific.

Why do you need a str cast here? Should not b be necessary

13/souhailshiv/Highest_Rated_Movie_Directors.py
```diff
@@ -0,0 +1,83 @@
...
+            if int(item[1]) >= MIN_YEAR:
```
pmayd, line 51: Don't use index access here, movie is a named tuple so you can access its fields with dot notation

13/souhailshiv/Highest_Rated_Movie_Directors.py
```diff
@@ -0,0 +1,83 @@
...
+
+        avrege.update({k: mean(moviescores)})
+
+    return avrege
+
+
+def print_results(avrege, directors):
+    avrege = {k: v for k, v in sorted(
+        avrege.items(), key=lambda item: item[1], reverse=True)}
+    i = 0
```
pmayd, line 62: Don't use i explicitly, use pythons enumerate function

13/souhailshiv/Highest_Rated_Movie_Directors.py
```diff
@@ -0,0 +1,83 @@
...
+    for k, v in avrege.items():
+        i = i+1
+        if i == 21:
+            break
+        print(f'\n{colored("%.2d"%i,"magenta")} {k}',
```
pmayd, line 67: This is hard to read, maybe a helper function for rendering directors

pmayd (approved):
Code is not very pythonic and hard to read, there are a lot of unnecessary type casts and for loops. You create to many dictionaries and don't use python constructs efficiently. Try to spend more time thinking about what you want to code instead of hacking a solution.

PR #774, opened 2020-10-06
Movies challenge

andreitch:
Difficulty level (1-10): [7]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Practicing fundamentals! ]

### DIFF
--- a/13/andreitch/directors.py
+++ b/13/andreitch/directors.py
@@ -1,0 +1,71 @@
+import csv
+from collections import defaultdict, namedtuple
+from functools import reduce
+
+MOVIE_DATA = '../movie_metadata.csv'
+NUM_TOP_DIRECTORS = 20
+MIN_MOVIES = 4
+MIN_YEAR = 1960
+
+Movie = namedtuple('Movie', 'title year score')
+
+
+def get_movies_by_director():
+    '''Extracts all movies from csv and stores them in a dictionary
+    where keys are directors, and values is a list of movies (named tuples)'''
+    directors = defaultdict(list)
+    with open(MOVIE_DATA, newline='') as csvfile:
+        reader = csv.DictReader(csvfile)
+        for row in reader:
+            if row['title_year'] and int(row['title_year']) >= MIN_YEAR:
+                directors[row['director_name']].append(
+                    Movie(row['movie_title'], row['title_year'], float(row['imdb_score'])))
+    return directors
+
+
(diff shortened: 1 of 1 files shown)


13/andreitch/directors.py
```diff
@@ -0,0 +1,71 @@
...
+                    Movie(row['movie_title'], row['title_year'], float(row['imdb_score'])))
+    return directors
+
+
+def get_average_scores(directors):
+    '''Filter directors with < MIN_MOVIES and calculate averge score'''
+    new_dict = {}
+    for k, v in directors.copy().items():
+        if len(v) >= MIN_MOVIES:
+            new_dict[(k, _calc_mean(directors[k]))] = v
```
pmayd, line 31: this is something you can do, but maybe should not do, its hard to maintain a dictionary with compound keys like this. You may come up with a different solution where you change the value to something like (score, movies), that is a Tuple with a float and a list

13/andreitch/directors.py
```diff
@@ -0,0 +1,71 @@
...
+    return new_dict
+
+
+def _calc_mean(movies):
+    '''Helper method to calculate mean of list of Movie namedtuples'''
+    return round(reduce(lambda a, b: a+b, [m.score for m in movies]) / len(movies), 1)
```
pmayd, line 37: it is ok to use functional programming and the reduce function, but I don't see the value here, I think it is easier to use `sum` and divide by the len? So `sum([m.score for m in movies])` should do the same?

13/andreitch/directors.py
```diff
@@ -0,0 +1,71 @@
...
+
+def print_results(directors):
+    '''Print directors ordered by highest average rating. For each director
+    print his/her movies also ordered by highest rated movie.
+    See http://pybit.es/codechallenge13.html for example output'''
+    fmt_director_entry = '{counter:02d}. {director:<52} {avg}'
+    fmt_movie_entry = '{year}] {title:<50} {score}'
+    sep_line = '-' * 60
+    report = sorted(directors.items(),
+                    key=lambda x: float(x[0][1]), reverse=True)
```
pmayd, line 48: this is what I meant, because you choose a data structure with compund dict keys, your lambda is now really a little complicated ;)

pmayd (approved):
Good work, try to use a more intuitive data structure that is easier to process

PR #792, opened 2021-02-07
PCC29 Geoff Riley

GeoffRiley:
Difficulty level (1-10): [4]
Estimated time spent (hours): [10]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [First time of playing with DJango, so it was all new in many ways; okay, the Python is familiar, but the way of implementing things in more of a Ruby kind of way felt strange; on the other hand it feels far more stable and logical than most of the php frameworks that I've used—and I include 'Cake' in that, although Cake has moved on a long way since I last used it and so it is probably much better now.
The only thing that actually foxed me was attempting to get syntax colouring to work when rendering some markdown text (yes, I went beyond the suggested ideas).]
Other feedback (what can we improve?): [I created a Django application called 'blog' and it took me absolutely ages to figure out why git was refusing to add it without forcing.  I tracked it down to the `.gitignore` file at the root of the repository excluding `**log`: what **is** missing from that file is an exclusion for `**sqlite3`.]


bbelderbos:
@GeoffRiley thanks, let me fix that!

PR #796, opened 2021-04-15
Pcc52

ttafsir:

Difficulty level (1-10): [3]
Estimated time spent (hours): [1/4]
Completed (yes/no): [yes]
I stretched my coding skills (if yes what did you learn?): [it was fun]

### DIFF
--- a/52/ttafsir/pomodoro.py
+++ b/52/ttafsir/pomodoro.py
@@ -1,0 +1,23 @@
+from rich.console import Console
+from rich.layout import Layout
+from rich.panel import Panel
+from rich.live import Live
+from time import sleep
+from datetime import timedelta
+
+
+console = Console()
+
+
+minutes = 25
+countdown_timer = timedelta(minutes=minutes)
+
+
+console.print(f"\nPOMODORO TIMER: {minutes}:00")
+with Live(auto_refresh=False) as live:
+    while countdown_timer:
+        console.print(str(countdown_timer), end="\r")
+        countdown_timer += timedelta(seconds=-1)
+        sleep(1)
+        live.refresh()
+    console.print("TIME IS UP!")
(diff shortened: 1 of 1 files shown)


52/ttafsir/pomodoro.py
```diff
@@ -0,0 +1,39 @@
...
+                self._countdown += timedelta(seconds=-1)
+                sleep(1)
+                live.refresh()
+
+    def restart(self, start_timer, prompt):
+        self._countdown = timedelta(minutes=start_timer)
+        self.run(prompt)
+
+
+timer_duration = int(input("Duration: ").strip())
```
bbelderbos, line 29: Put this into an `if __name__ == "__main__"` so it doesn't run if you would want to import this module.

52/ttafsir/pomodoro.py
```diff
@@ -0,0 +1,39 @@
+from rich.console import Console
```
bbelderbos, line 1: `rich`?

bbelderbos (approved):
NIce

52/ttafsir/pomodoro.py
ttafsir: I've been playing with `rich` lately and it seems super cool and useful for CLI outputs and layouts.

52/ttafsir/pomodoro.py
ttafsir: ah, good point. will do that.

ttafsir: Thanks!

52/ttafsir/pomodoro.py
bbelderbos: cool, need to check that out, maybe we can do an article on the blog ...

PR #812, opened 2021-10-01
PCC45 GeoffRiley

GeoffRiley:
Difficulty level (1-10): [4]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Brand new BDD system in `mamba` and `expects`… I've used similar tools in Ruby so it was good to use.  Only one problem uncovered that `expects` has a slight problem with it's `raise_error` detection.]


bbelderbos:
@GeoffRiley interesting! I want to look at `Behave` soon ...

PR #831, opened 2022-01-31
Pcc52

kaijuking:
Difficulty level (1-10): [5]
Estimated time spent (hours): [20]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Learned to use PySimpleGUI, dabbled with threads (though didn't use them in the final product), got more practice with datetime!]


52/kaijuking/pomodoro.py
```diff
@@ -0,0 +1,146 @@
...
+    def process_user_input(user_input: str):
+        if not user_input.isdigit():
+            return False
+        elif not (int(user_input) > 0 and int(user_input) <= 120):
+            return False
+        else:
+            return True
+
+
+    while True:
```
bbelderbos, line 63: I would break this down into more functions or methods

52/kaijuking/pomodoro.py
```diff
@@ -0,0 +1,146 @@
+from datetime import datetime, timedelta
+import PySimpleGUI as sg
+
+
+def pomodoro_gui():
```
bbelderbos, line 5: Which so much logic embedded and possible need to keep state I would make this a class

bbelderbos (approved):
Nice work

PR #843, opened 2022-06-23
My solution to PCC01

thechrislee:
Difficulty level (1-10): [2]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [Not necessarily coding, but this is probably the most I've ever done with git besides cloning a repo. I've never really branched or worked with repos that have other contributors.
This challenge is the similar if not the same as one of the pybites challenges that I already completed. Truthfully, the hardest part is getting familiar with git. I need more practice.]
Other feedback (what can we improve?): [Rewrite the tests to use pytest...I may do that myself. Please cleanup the repo. When I switched to the community branch, I noticed a a file with the answers in the root of the 01 challenge. ]


01/thechrislee/wordvalue.py
```diff
@@ -0,0 +1,24 @@
...
+
+def load_words(path=DICTIONARY):
+    """Load dictionary into a list and return list"""
+    with open(path) as file:
+        return [line.strip() for line in file]
+
+def calc_word_value(word):
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    total = 0
```
bbelderbos, line 11: can use `sum()` built-in

bbelderbos (approved):
Nice

PR #844, opened 2022-09-24
Pcc45

mhered:

Difficulty level (1-10): [2]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [learned some TDD]

### DIFF
--- a/45/mhered/app/fizzbuzz.py
+++ b/45/mhered/app/fizzbuzz.py
@@ -1,0 +1,17 @@
+#!/usr/bin/python3
+
+
+def fizzbuzz(num: int):
+    """
+    FizzBuzz(int) plays FizzBuzz
+    """
+    if not isinstance(num, int):
+        raise ValueError("argument should be of type int")
+    else:
+        if num % 5 == 0 and num % 3 == 0:
+            return "FizzBuzz"
+        if num % 5 == 0:
+            return "Buzz"
+        if num % 3 == 0:
+            return "Fizz"
+        return num
--- a/45/mhered/test/test_fizzbuzz.py
+++ b/45/mhered/test/test_fizzbuzz.py
@@ -1,0 +1,39 @@
+#!/usr/bin/python3
+
+import unittest
+
+from app.fizzbuzz import fizzbuzz
+
+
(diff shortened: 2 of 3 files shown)


45/mhered/test/test_fizzbuzz.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def test_fizzbuzz_returns_ints(self):
+        samples = [1, 2, 4, 6, 7, 8, 11, 13]
+        for sample in samples:
+            result = fizzbuzz(sample)
+        self.assertEqual(result, sample)
+
+    def test_fizzbuzz_returns_fizz_with_multiples_of_3(self):
+        samples = [3, 6, 9, 12]
+        for sample in samples:
```
bbelderbos, line 18: You can use paremetrize: https://docs.pytest.org/en/6.2.x/parametrize.html

45/mhered/test/test_fizzbuzz.py
```diff
@@ -0,0 +1,39 @@
...
+            result = fizzbuzz(sample)
+        self.assertEqual(result, "Buzz")
+
+    def test_fizzbuzz_returns_fizzbuzz_with_multiples_of_3_and_5(self):
+        samples = [15, 30, 45, 60]
+        for sample in samples:
+            result = fizzbuzz(sample)
+        self.assertEqual(result, "FizzBuzz")
+
+    def test_fizzbuzz_raises_error_message_if_arg_not_int(self):
```
bbelderbos, line 34: nice descriptive name, I don't mind test function names to be longer

45/mhered/app/fizzbuzz.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/python3
+
+
+def fizzbuzz(num: int):
+    """
+    FizzBuzz(int) plays FizzBuzz
+    """
+    if not isinstance(num, int):
```
bbelderbos, line 8: mypy would also catch this, but it does not prevent the function from being used by other types, after a `raise` you don't need the else so you could dedent the code below one level

PR #845, opened 2022-10-05
PCC01 NNRepos

NNRepos:
Difficulty level (1-10): [1]
Estimated time spent (hours): [0.5]
Completed (yes/no): [yes]
Other feedback (what can we improve?): [probably makes more sense to use relative imports, 
  that way we don't have to copy and delete the files to run the tests.]

### DIFF
--- a/01/NNRepos/wordvalue.py
+++ b/01/NNRepos/wordvalue.py
@@ -1,0 +1,23 @@
+from data import DICTIONARY, LETTER_SCORES
+from typing import Optional, List
+
+
+def load_words() -> List[str]:
+    """Load dictionary into a list and return list"""
+    with open(DICTIONARY) as f:
+        return [line.strip() for line in f.readlines()]
+
+
+def calc_word_value(word: str) -> int:
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    return sum(LETTER_SCORES[letter] for letter in list(word.replace('-', '').upper()))
+
+
+def max_word_value(dictionary: Optional[List[str]] = None) -> str:
+    """Calculate the word with the max value, can receive a list
+    of words as arg, if none provided uses default DICTIONARY"""
+    if dictionary is None:
+        dictionary = load_words()
+    print("crap:", [x for x in dictionary if "-" in x])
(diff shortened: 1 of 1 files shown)


01/NNRepos/wordvalue.py
```diff
@@ -0,0 +1,23 @@
...
+    return sum(LETTER_SCORES[letter] for letter in list(word.replace('-', '').upper()))
+
+
+def max_word_value(dictionary: Optional[List[str]] = None) -> str:
+    """Calculate the word with the max value, can receive a list
+    of words as arg, if none provided uses default DICTIONARY"""
+    if dictionary is None:
+        dictionary = load_words()
+    print("crap:", [x for x in dictionary if "-" in x])
+    return list(sorted(([calc_word_value(word), word] for word in dictionary), reverse=True))[0][1]
```
bbelderbos, line 23: `sorted()` already returns a list. I would break this out into multiple lines, which would make it more readable.

01/NNRepos/wordvalue.py
```diff
@@ -0,0 +1,23 @@
...
+def load_words() -> List[str]:
+    """Load dictionary into a list and return list"""
+    with open(DICTIONARY) as f:
+        return [line.strip() for line in f.readlines()]
+
+
+def calc_word_value(word: str) -> int:
+    """Calculate the value of the word entered into function
+    using imported constant mapping LETTER_SCORES"""
+    return sum(LETTER_SCORES[letter] for letter in list(word.replace('-', '').upper()))
```
bbelderbos, line 14: I don't think you need the `list()`, `word` is already iterable. The solution also uses `LETTER_SCORES.get(char.upper(), 0)` which is safer, because with `LETTER_SCORES[letter]` any unexpected letter would raise a `KeyError`.

bbelderbos (approved):

01/NNRepos/wordvalue.py
NNRepos: i actually used the `KeyError` for debugging (that's how i've discovered that `-` is a legal character), but for correctness, i guess you're right.

01/NNRepos/wordvalue.py
NNRepos: i was actually looking for something like this: `key=calc_word_value`. that's very nice

PR #847, opened 2022-10-08
PCC02 NNRepos

NNRepos:
notes:
* i ignored internal implementation tests
* also improved PCC01

Difficulty level (1-10): [2]
Completed (yes/no): [yes]
Other feedback (what can we improve?): [i don't think tests should be testing internal implementation. a good test only tests behaviour. that's why i did not create some of the required implementation methods (e.g. `_validate`).]

### DIFF
--- a/02/NNRepos/game.py
+++ b/02/NNRepos/game.py
@@ -1,0 +1,66 @@
+#!python3
+# Code Challenge 02 - Word Values Part II - a simple game
+# http://pybit.es/codechallenge02.html
+
+import random
+from collections import Counter
+from typing import Dict, List
+
+from data import DICTIONARY, LETTER_SCORES, POUCH
+
+NUM_LETTERS = 7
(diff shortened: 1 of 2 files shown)


02/NNRepos/game.py
```diff
@@ -0,0 +1,66 @@
...
+    return max(words, key=calc_word_value)
+
+
+def draw_letters() -> List[str]:
+    return random.sample(POUCH, NUM_LETTERS)
+
+
+def is_legal_word(word: str, drawn_letters_count: Dict[str, int]) -> bool:
+    word_letters_count = Counter(word.upper())
+    for letter, count in word_letters_count.items():
```
bbelderbos, line 33: Could also use `any`/ `all` built-ins

NNRepos: that's right, didn't think about it :)

PR #871, opened 2023-07-19
Finished

zambam5:
Difficulty level (1-10): [1]
Completed (yes/no): [Yes]
I stretched my coding skills (if yes what did you learn?): [I chose to figure out how to make a notification sound when the timer ends, something I had never considered doing before.]

### DIFF
--- a/52/zambam5/timer.py
+++ b/52/zambam5/timer.py
@@ -1,0 +1,49 @@
+import platform
+
+from datetime import datetime, timedelta
+from time import sleep
+
+if platform.system() == "Windows":
+    from winsound import Beep
+
+
+def pomodoro_timer(duration: int = 20, beep: bool = False):
+    """
+    Take a duration and begin a timer loop
+    """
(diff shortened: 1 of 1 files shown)


52/zambam5/timer.py
```diff
@@ -0,0 +1,49 @@
...
+        else:
+            beep = True
+    else:
+        beep = False
+    while True:
+        pomodoro_timer(beep=beep)
+        again = input(
+            'Do you want to start the timer again? Answer "yes" or "no" only '
+        )
+        if again == "yes":
```
pmayd, line 40: You don't have to check here, you can completely omit this branch, only check for when you want to do something. Here you want to do something if the user does not want to continue so only check for "no"

pmayd (changes requested):
Solid work, but room for improvement and refactoring :)

pmayd (approved):
Solid work but room for improvement:)

